Issue #7, opened 2021-12-03
v0.8 test fixtures


expede:
Generate valid and invalid example UCANs to test implementations against. At least one per failure type (e.g. rights escalation), and one valid per key type and feature (root UCAN, single vs multiple capabilities, and so on).

The Haskell UCAN codebase can generate random UCANs on the fly (via QuickCheck) so it may be easiest to generate them there.

expede:
@appcypher I saw you asking where the fixtures lived on Discord. Because we have multiple implementations, we create fixtures themselves in the `ucan-wg/spec` repo (e.g. [`tests/0.7` fixtures](https://github.com/ucan-wg/spec/tree/main/tests/0.7.0)), labelled by how they're broken (or not broken, depending). Implementers can then write tests against these.

appcypher:
@expede The Haskell UCAN codebase you mentioned above, is it [fission/hs-ucan](https://github.com/fission-suite/fission/tree/main/hs-ucan)?

expede:
@appcypher it is, yes. @matheus23 currently has the hood up on it, but it will move to `ucan-wg` in the near future :)

expede:
Probably worth mentioning that `hs-ucan` will generate _valid_ UCANs only, which you'd need to "break" by hand for test cases.

PR #8, opened 2022-01-04
Add initial tests

bgins:
## Summary

This PR implements the following **features**

* [x] Add test cases for UCAN version 0.7.0

This PR adds an initial set of test cases that UCAN implementations can test against. It includes test cases for:

- Valid UCANs, some with valid proofs
- Expired and too early UCANs
- Invalid signature
- UCAN issuer does not match proof audience
- Invalid proofs

Note that the tests do not explicitly test attenuations, but the attenuations in the test cases are kept simple and should be valid.

## Test plan (required)

The test cases should be run against up to date, existing UCAN implementations. I have tested these with the `ts-ucan` library.

## Closing issues

We may want to add more test cases, but this gets us started on #7.


expede (approved):
Looking awesome @bgins!

PR #20, opened 2022-02-04
v0.8.1 — Text clarifications and `as:` ability

### LINKED ISSUE #3: Clarify invocation flow
Clarified this live, but should be in the spec directly https://github.com/ucan-wg/ts-ucan/issues/17

### LINKED ISSUE #15: Image uses "potency" instead of "ability"
https://github.com/ucan-wg/spec/blob/9efe6b7c490dfcd1246267796e34dc27e95f0c29/README.md?plain=1#L61

This image uses the term "potency", but we've switched to being consistent on "ability"

expede:
Closes #3 
Closes #15 
Closes https://github.com/ucan-wg/ts-ucan/issues/49


README.md
```diff
@@ -247,9 +247,11 @@ The `can` field describes the verb portion of the capability: an action that can
 
 Abilities MAY be organized in a hierarchy with enums. A common example is superuser access ("anything") on a file system. Another is read vs write access, such that in an HTTP context `WRITE` implies `PUT`, `PATCH`, `DELETE`, and so on. Organizing potencies this way allows for adding more options over time in a backwards-compatible manner, avoiding the need to reissue UCANs with new resource semantics.
 
-Abilities MUST NOT be case sensitive, and MUST be namespaced by at least one path segment. For instance, `http/PUT` and `foo/PUT` MUST be treated as unique from each other.
+Abilities MUST NOT be case sensitive. For example, `http/post`, `http/POST`, `HTTP/post`, `HTTP/POST`, and `hTtP/pOsT` MUST all mean the same ability.
```
expede, line 250: wHy YeS, wE sUpPoRt SaRcAsM cAsE

README.md
```diff
@@ -286,13 +288,22 @@ Proofs referenced by content address MUST be resolvable by the recipient, for in
...
+
+This is useful in several cases, for example:
+
+1. When delegating all resources, like in a [`my` scheme](#4.2-all-owned-resources)
+2. To grant the maximum ability when the current ability semantics may be extended later
+
+## 4.2 All Owned Resources
 
-### 4.1.1. `my` Scheme
+### 4.2.1. `my` and `as` Schemes
```
expede, line 302: This was in an earlier draft, but then I tried to cover it with `my` since they're pretty similar. It's much clearer this way, so I'm glad that Gozala spoke up!

README.md
```diff
@@ -348,15 +382,27 @@ If any of the following criteria are not met, the UCAN MUST be considered invali
 
 A UCAN's time bounds MUST NOT be considered valid if the current system time is prior to the `nbf` field, or after the `exp` field. This is called "ambient time validity".
 
-All witnesses MUST contain time bounds equal to or wider than the UCAN being delegated to. If the witness expires before the outer UCAN — or starts after it — the reader MUST treat the UCAN as invalid. This is called "timely delegation".
+All witnesses MUST contain time bounds equal to or wider than the UCAN being delegated to. If the witness expires before the outer UCAN — or starts after it — the reader MUST treat the UCAN as invalid. This is called "timely delegation". This conditions MUST hold even if the current wall clock time is inside the misdelegated bounds.
```
QuinnWilton, line 385: Typo: "This conditions" -> "These conditions"

README.md
```diff
@@ -348,15 +382,27 @@ If any of the following criteria are not met, the UCAN MUST be considered invali
...
 A UCAN is valid inclusive from the `nbf` time, and until the `exp` field. If the current time is outside of these bounds, the UCAN MUST be considered invalid. A delegator or invoker SHOULD account for expected clock drift when setting these bounds. This is called "timely invocation".
 
 ## 5.2 Principal Alignment
 
 In delegation, the `aud` field of every witness MUST match the `iss` field of the outer UCAN (the one being delegated to). This alignment MUST form a chain all the way back to the originating principal for each resource.
 
-An agent discharging a capability MUST verify that the outermost `aud` field matches its own DID. If they do not match, the associated action MUST NOT be performed.
+### 5.2.1 Invocation Recipient Validation
+
+An agent discharging a capability MUST verify that the outermost `aud` field _matches its own DID._ If they do not match, the associated action MUST NOT be performed. This is REQUIRED in order to prevent the misuse of UCANs in the unintended context.
```
QuinnWilton, line 395: Should this be: "to prevent the misuse of UCANs in an unintended context."?

README.md
```diff
@@ -348,15 +382,27 @@ If any of the following criteria are not met, the UCAN MUST be considered invali
...
+
+The following UCAN fragment would be be valid to dischange as `did:key:zH3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV`. The agent with DID `did:key:z6MkiTBz1ymuepAQ4HEHYSF1H8quG5GLVVQR3djdX3mDooWp` MUST NOT accept this UCAN.
```
QuinnWilton, line 397: Typo: "be be" -> "be"

README.md
expede: Thanks!

)I'm going to run this through a spell checker before flipping it out of Draft)

QuinnWilton: Whoops, sounds good! I was taking a break from reading and saw the notifications came in so I took a quick look :)

expede: No problem — seriously thank you! Mainly there's going to be a bunch of these, and we can let a computer fix most of my late night typos 😛 I would love a proof read when it's ready, though!

QuinnWilton: Sounds good! Mostly I'm trying to stay on top of the changes because each iteration gives me more context about the spec, and forms a clearer picture of it in my head

README.md
```diff
@@ -76,32 +76,59 @@ For instance, `wnfs/APPEND` is an action for WebNative filesystem paths. The act
...
+The "pool" is the total rights of the authorization space down to the relevant volume of authorizations. Individual capabilities MAY overlap; the pool is the union. With the exception of [rights amplification](#54-rights-amplification), every individual delegation MUST have equal or narrower capabilties from their delegator. Inside this content space, you can draw a boundary around some resource(s) (their type, identifiers, and paths or children), and their capabilities.
 
-As a practical matter, since scopes form a group, you can be fairly loose: order doesn’t matter, and merging resources can be quite broad since the more capable of any overlap will take precedence (i.e. you don’t need a clean separation).
+As a practical matter, since pools form a group, you can be fairly loose: order doesn’t matter, and merging resources can be quite broad since the more capable of any overlap will take precedence (i.e. you don’t need a clean separation).
```
expede, line 110: @QuinnWilton

_This_ I could use a hand with while WIP, though! The feedback is that "scope" is confusing in a programming context (we use that word a LOT). It does really capture the idea, and I'm struggling to find a replacement that really fits.

It could be "capability set", but that doesn't feel very descriptive. "Pool" came up because I had a side conversation with Philipp about worker pools, and it's amorphous enough, but still doesn't really capture the intent.

Any thoughts?

expede:
Closes #15

README.md
expede: 🤔 "Collection"

expede: `<mumbles something about the wallet metaphor being terrible>`

QuinnWilton: How do you feel about "capability space"? I've been visualizing the capabilities carried by a UCAN as forming a 3D space with resources, actions, and identities along each axis, which would mean that the capabilities represented by the union of two delegations would be given by the union of their capability spaces.

QuinnWilton: I suppose you already use the term "authorization space" here, so you already get into that idea a bit, but I am a big fan of the geometric interpretation, so something along those lines might be helpful: "capability boundary", "capability sphere", etc

QuinnWilton: "capability sphere" doesn't quite feel right, because that implies a symmetry that doesn't necessarily exist. I think "capability boundary" is really the closest name for what I'm visualizing, if you think of it from a topological sense.

QuinnWilton: Alright, I think I've traced some of my confusion to the use of "paths or children" in the above passage. I'm guessing that's referring specifically to Webnative, so I'm going to make some assumptions here to clarify my understanding of things.

It's possible that all of this is entirely wrong though, because my interpretation of this spec is mostly based on this document and some educated guesses. If not, I'm no longer convinced that a geometric interpretation is helpful, and I'll to explain why by considering WebNative paths and email addresses. I'll assume that paths form a tree, and that a capability against some path also applies to all of that path's descendants.

I'll also assume that we have four actions:
```
wnfs/APPEND
wnfs/OVERWRITE (implies wnfs/APPEND)
wnfs/READ
email/SEND
```

Some filesystem:
```
Private/
├─ Pictures/
├─ Documents/
```

Some email addresses:
```
alice@example.com
bob@example.com
```

And two capabilities:

```
A = wnfs/APPEND × "Private/"
B = wnfs/OVERWRITE × "Private/Documents/"
```

Then:
- `A` grants `wnfs/APPEND` against the entire filesystem
- `B` grants `wnfs/APPEND` and `wnfs/OVERWRITE` against `"Private/Documents/"`
- `wnfs/READ` has not been granted against any paths
- `email/SEND` has not been granted against any email addresses

The spec already mentions that actions can take on any semantics, but this example shows that the same is true for resources. The paths form a partial order, over the "descendant of" relationship, but there's no way to compare paths that aren't descendants. Similarly, `{ wnfs/APPEND, wnfs/OVERWRITE }` forms a partial order, but there's no relationship between email addresses whatsoever, email addresses and paths, and `wnfs/READ` or `email/SEND` with any other actions.

All of this together means that we don't have one connected space to represent all of the possible capabilities in, and instead have a disconnected space made up of three subspaces:
1) Write capabilities against filesystem paths, with points in `{ NONE, APPEND, OVERWRITE }`
2) Read capabilities against filesystem paths, with points in `{ NONE, READ }`
3) Send capabilities against email addresses, with points in `{ NONE, SEND }`

If I'm understanding your description of things correctly, then what you're tentatively calling a "capability pool" refers to one of these subspaces. Which is to say that there is a way to make a geometric / topological interpretation of this stuff work, but I think it's too confusing to be elucidating, and requires far too much knowledge of the underlying semantics of the resources and the actions to easily visualize.

Maybe this can give us a starting point for thinking about clearer language though? At the very least, this leaves me feeling like what you're against calling a "capability set" is likely more accurately described as a "capability poset", though I don't think that's a great term either.

My favourite term is still "capability boundary", but I'm wary of the complexity involved in conveying the motivations behind the term. That being said, since posets can be viewed as trees, maybe the problem can shifted toward one of viewing the authorization space as a forest of capabilities?

Then a "capability pool" would become a "capability tree", and would compose under the semantics given by the underlying `(resource, action)` pair. My main reservation here comes from not knowing whether that structure is guaranteed to be a poset, or if richer structures like lattices are possible. I'm assuming not, because all of this is being built on top of IPFS, but I want to raise the concern!

The tree interpretation is still pretty abstract, but maybe with enough concrete examples it would be a more natural perspective on the problem, given that graphs are likely more familiar to most engineers than topological spaces?

Anyway, I have no idea if this is all gibberish or not, but I'm happy to chat with you in more detail about it either way!

expede: > How do you feel about "capability space"?

I tend to use this as "the complete set of all possible capabilities", in the same sense as "the solution space". But maybe that's just me?

expede: >  but this example shows that the same is true for resources

Yup! They're literally URIs, so however that URI works is their semantics. Perhaps good to spell that out explicitly 🤔 👍

expede: >  I worry that it's too confusing to be elucidating, and requires too much knowledge of the underlying semantics of the resources and the actions to easily visualize.

Oh, you're going way deeper! I want a term that captures the idea of the union of all capabilities that are currently available to the current agent (whoever has its DID) that are provided by this UCAN. It's a set union on all of their capabilities, but I don't want to use the phrase "set union over capabilities" because it's overly technical and wordy.

The other thing that I'm trying to capture is the concept of overlapping. For example, we could have the following:

```
"wnfs:/.../pictures" x WRITE
"wnfs:/.../pictures/vacation" x DESTROY
```

Where DESTROY > WRITE. These overlap, but not perfectly (to say nothing of overlapping paths in a DAG).

What I liked about "scope" was that it's essentially your Γ. Everything that's in your execution environment with the UCAN loaded, which is to say "the current scope".

expede: > not knowing whether those semantics are guaranteed to be a poset

They are! We use posets in the code during validation.

_Each_ capability forms a DAG of capability "strength" or "size" (to use the visual metaphor). Technically having several is like an ad hoc version of that same idea (literally an algebraic product type). But! I don't want to scare people off with lots of math.

expede: (You see how I'm struggling here? 😅😅😅 Why is naming things so hard?!)

bgins: My opinion is that scope is the right word. Maybe changing that section title to “Scope of Authority” would help to introduce the type of scope being discussed.

I looked at a couple of synonyms like “purview” or “ambit”, but those don’t make sense in a programming context. 😆

QuinnWilton: > > How do you feel about "capability space"?
> 
> I tend to use this as "the complete set of all possible capabilities", in the same sense as "the solution space". But maybe that's just me?

I agree with this!

> > but this example shows that the same is true for resources
> 
> Yup! They're literally URIs, so however that URI works is their semantics. Perhaps good to spell that out explicitly 🤔 👍

I would lean toward being explicit, but I don't think you need to get too deep about it. Just a quick sentence like "Each resource MAY have its own semantics. For example, WebNative filesystem paths form a hierarchy, and can be visualized as a tree." would make explicit that the paths can behave more like a tree than like opaque identifiers.

> What I liked about "scope" was that it's essentially your Γ. Everything that's in your execution environment with the UCAN loaded, which is to say "the current scope". You don't really need to think about the interaction so much, since they're all phrased in the positive, "any witness will do"

This is a really good insight, and maybe the way to express this is just by connecting back to section 1.3: "Each UCAN includes a constructive set of assertions of what it is allowed to do. Note that this is not a predicate: it is a positive assertion of rights."

Having thought through this conversation, I think I agree with you and @bgins about scope being a really good word to use here, so maybe the terms themselves don't actually need to change, and you just need to define them upfront and give an example? Something like:

> The set of capabilities granted by a UCAN is called its scope. Scopes give a declarative description of the granted abilities, and two scopes can be combined by concatenating all of their capabilities together. Since broader capabilities absorb weaker capabilities, this process is additive, and scopes can be combined in any order, with the result always being at least as broad as each of the original scopes.
>
> For example, given the following scopes against a WebNative filesystem, they can be merged as follows:
> ```
> ScopeA = { "wnfs:/.../pictures" x WRITE }
> ScopeB = { "wnfs:/.../pictures/vacation" x DESTROY } 
> 
> merge(ScopeA, ScopeA) => {
>    "wnfs:/.../pictures" x WRITE,
>    "wnfs:/.../pictures/vacation" x DESTROY
> }
> ```
> This new scope partially overlaps, and provides both `WRITE` and `DESTROY` capabilities against `"wnfs:/.../pictures/vacation"`. 

Do you think a description like that might address any of the confusion?

> They are! We use posets in the code during validation.

Awesome, that's good to know! I agree about not wanting to overload readers with any of the theory, but that's very helpful for my own understanding of things. Thanks for bearing with me while I wrapped my head around things!

QuinnWilton: > (You see how I'm struggling here? 😅😅😅 Why is naming things so hard?!)

Something something names are the limit of our world something something

QuinnWilton: If you don't want to use the word "scope", then you might be able to substitute "domain"? I'm reminded of realms in [RFC 2617](https://datatracker.ietf.org/doc/html/rfc2617#section-1.2):

> The realm value, in combination with the canonical root URL of the server being accessed, defines the protection > space.
> ...
> The protection space determines the domain over which credentials can
> be automatically applied.

It's not a perfect analogy, but there may be something to steal from there?

bgins: > > The set of capabilities granted by a UCAN is called its scope. Scopes give a declarative description of the granted abilities, and two scopes can be combined by concatenating all of their capabilities together. Since broader capabilities absorb weaker capabilities, this process is additive, and scopes can be combined in any order, with the result always being at least as broad as each of the original scopes.
> > For example, given the following scopes against a WebNative filesystem, they can be merged as follows:
> > ```
> > ScopeA = { "wnfs:/.../pictures" x WRITE }
> > ScopeB = { "wnfs:/.../pictures/vacation" x DESTROY } 
> > 
> > merge(ScopeA, ScopeA) => {
> >    "wnfs:/.../pictures" x WRITE,
> >    "wnfs:/.../pictures/vacation" x DESTROY
> > }
> > ```
> > This new scope partially overlaps, and provides both `WRITE` and `DESTROY` capabilities against `"wnfs:/.../pictures/vacation"`.

This explanation using scope reads great to me! :clap:

expede:
NOTE TO SELF:

>   "challenge": sign(did.privateKey, `REVOKE:${ucanCid}`)

Should specify the hash function for the signature

README.md
```diff
@@ -332,11 +428,24 @@ prf = "prf" selector
 selector = "*" / 1*DIGIT
 ```
 
-`prf:*` represents all of the UCANs in the current proof scope. The witnesses in the `prf` field may be referenced by their index, starting at 0: `prf:0`
+`prf:*` represents all of the UCANs in the current proofs array. The witnesses for the current UCAN MAY be referenced by their index in the `"prf"` field. If selecting a particular witness (i.e. not the wildcard), then a zero-indexed MUST be used. The first UCAN would be selected by `prf:0`, the second by `prf:1`, and so on. By virtue of the indexing scheme, selections MUST be performed on the current UCAN only, and cannot recursively on nested witnesses.
 
-### 4.2.2 `prf` Actions
+Further selection of capabilities inside fo specific witnesses MUST NOT be a valid parsing of this URI. For instance, `prf:0:mailto` MUST NOT be a valid `prf` URI.
+
+### 4.3.2 `prf` Actions
```
matheus23, line 435: Re-reading this I'm wondering whether we can get away with only the `as:` *or* the `prf:` scheme, as they seem like they could both be used for e.g. device linking, although `prf:*` allows compacting UCAN delegations.

(I feel like I'm arguing against my past self)

I'm clarifying my thinking as I'm writing this.
`as:did:key:zMABC` must only be selecting all ambient authority of a certain DID, but *not* select the capabilities that the DID has in the particular UCAN chain,  or does it?

matheus23 (approved):

README.md
```diff
@@ -348,21 +457,37 @@ If any of the following criteria are not met, the UCAN MUST be considered invali
...
 ## 5.2 Principal Alignment
 
 In delegation, the `aud` field of every witness MUST match the `iss` field of the outer UCAN (the one being delegated to). This alignment MUST form a chain all the way back to the originating principal for each resource.
 
-An agent discharging a capability MUST verify that the outermost `aud` field matches its own DID. If they do not match, the associated action MUST NOT be performed.
+### 5.2.1 Invocation Recipient Validation
+
+An agent discharging a capability MUST verify that the outermost `aud` field _matches its own DID._ If they do not match, the associated action MUST NOT be performed. This is REQUIRED in order to prevent the misuse of UCANs in an unintended context.
+
+The following UCAN fragment would be valid to dischange as `did:key:zH3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV`. Any other agent  MUST NOT accept this UCAN. For example, `did:key:z6MkiTBz1ymuepAQ4HEHYSF1H8quG5GLVVQR3djdX3mDooWp` MUST NOT discharge it.
```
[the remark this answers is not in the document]
expede: Ah, yes, sorry this is technical jargon. I'll rephrase it. Here it means to perform the action associated with the capability.

README.md
expede: You're right that these overlap a bit . The challenge with `prf` is that it doesn't let you reduce scope; it's really there to save on space so the thinking is that it should be as direct as possible and any use off of that would need to use some explicit form. So, if someone delegates all of their resources to me, I can't redelegate just all email addresses, for example.

Any thoughts?

README.md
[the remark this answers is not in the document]
matheus23: I think it was also just a typo "dischange" vs. "discharge", but I like the new version :smile: :+1:

README.md
```diff
@@ -208,16 +283,16 @@ The OPTIONAL `fct` field contains arbitrary facts and proofs of knowledge. The e
...
-### 3.2.5 Attenuation Scope
+### 3.2.5 Attenuations
 
-The attenuation scope (i.e. UCAN output, or "caveats") MUST be an array of heterogeneous access scopes (defined below). This array MAY be empty.
+The attenuations (i.e. UCAN output, or "caveats") MUST be an array of heterogeneous capabilities (defined below). This array MAY be empty.
 
 This array MUST contain some or none of the following:
 1. A strict subset (attenuation) of the witnesses
 2. Witnesses originated by the `iss` DID (i.e. by parenthood)
 3. Witnesses composed from other witnesses (see rights amplification)
```
matheus23, line 293: "Witness" in this context seems weird to me. I thought witness is the stuff inside the `prf` array. (I.e. other encoded UCANs).
I'd call them "Capabilities originated by the `iss` DID (i.e. by parenthood)" and "Capabilities composed from other capabilities in the witnesses (see rights amplification)".

README.md
```diff
@@ -332,11 +428,24 @@ prf = "prf" selector
 selector = "*" / 1*DIGIT
```
matheus23, line 428: I can't review the line above :(

But the line above should say `prf = "prf:" selector`. (It's missing the `:` in the ABNF)

matheus23 (commented):
I decided that reading spec diffs makes it really hard to review, so I had another run through the spec, but looked at the end result instead.
Now I'm commenting on stuff that wasn't changed in this PR :sweat_smile:

README.md
matheus23: Oh I totally agree about disallowing `prf:0:mailto` and similar.

I was mostly thinking about the complexity vs. optimization gain trade-off.
But thinking about this again, it's not that complex at all.

README.md
expede: @matheus23 better?

expede: https://github.com/ucan-wg/spec/pull/20/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R290-R293

README.md
```diff
@@ -76,32 +100,83 @@ For instance, `wnfs/APPEND` is an action for WebNative filesystem paths. The act
 
 A capability is the association of an action to a resource: `resource x action`
 
-## 2.4 Scope
+## 2.4 Capability Scope
+
+The set of capabilities delegated by a UCAN is called its "capability scope", often shortened to "scope". This functions as a declarative description of delegated abilities.
+
+Mergeing capability scopes MUST follow set semantics, where the result includes all capabilities from the input scopes. Since broader capabilities automatically include narrower ones, this process is always additive. Capability scopes can be combined in any order, with the result always being at least as broad as each of the original scopes.
```
bgins, line 107: > ~~Mergeing~~ Merging capability scopes MUST follow set semantics

Typo. Merging, not Mergeing.

README.md
```diff
@@ -76,32 +100,83 @@ For instance, `wnfs/APPEND` is an action for WebNative filesystem paths. The act
...
-The "scope" is the total rights of the authorization space down to the relevant volume of authorizations. With the exception of [rights amplification](#54-rights-amplification), every individual delegation MUST have equal or less scope from their delegator. Inside this content space, you can draw a boundary around some resource(s) (their type, identifiers, and paths or children), and their capabilities.
+ScopeB = [
+  { "with": "wnfs://alice.example.com/pictures/vacation/", "can": "wnfs/APPEND" };
+  { "with": "wnfs://alice.example.com/pictures/vacation/hawaii/", "can": "wnfs/OVERWRITE"}
+];
 
-As a practical matter, since scopes form a group, you can be fairly loose: order doesn’t matter, and merging resources can be quite broad since the more capable of any overlap will take precedence (i.e. you don’t need a clean separation).
+merge(ScopeA, ScopeA) == [
+   {"with": "wnfs://alice.example.com/pictures/", "can": "wnfs/APPEND"},
+   {"with": "wnfs://alice.example.com/pictures/vacation/", "can": "wnfs/OVERWRITE"}
```
bgins, line 153: Should `OVERWRITE` extend to all vacation pictures after merging? Before the merge, `ScopeB` allowed overwriting Hawaii pictures, but not all vacation pictures.

README.md
```diff
@@ -76,32 +100,83 @@ For instance, `wnfs/APPEND` is an action for WebNative filesystem paths. The act
...
+Delegation is the act of granting another principal (the delegate) the capability to use a resource that another has (the delegator). This MUST be proven by a "witness", which is either the signature of the owning principal, or a UCAN that has access to that capability in its capability scope.
 
-Each direct delegation leaves the scope of the action at the same level, or diminishes it. The only exception is in "rights amplification", where a delegation MAY be proven by one-or-more witnesses of different types, if part of the resource's semantics. 
+Each direct delegation leaves of the action at the same level, or diminishes it. The only exception is in "rights amplification", where a delegation MAY be proven by one-or-more witnesses of different types, if part of the resource's semantics. 
+
+Note that delegation is a seperate concept from invocation [§2.8](#28-invocation). Delegation is the act of granting a capability to another, not the act of using it (invocation) which has additional requirements.
```
bgins, line 164: Love the section mark and link here! ❤️

README.md
```diff
@@ -76,32 +100,83 @@ For instance, `wnfs/APPEND` is an action for WebNative filesystem paths. The act
...
 
 ## 2.7 Revocation
 
 Revocation is the act of invalidating a UCAN after the fact, outside of the limitations placed on it by the UCAN's own fields (such as its expiry). 
 
 In the case of UCAN, this MUST be done by a witness issuer DID. For more on the exact mechanism, see the revocation validation section.
 
+## 2.8 Invocation
+
+UCANs are used to delegate capabailities between DID-holding agents, eventually terminating in an "invocation" of those capabilities. This is when the capability is exercised to perform some task on the resource. Note that **the only agent that is allowed to perform an action with a UCAN MUST be the one holding the DID private key associated with the `aud` field**. For more on the specifics of this validation, see [§5.2.1](#521-invocation-recipient-validation).
```
bgins, line 178: The "This" opening the second sentence could be made more concrete by repeating invocation. Something like:

> Invocation occurs when the capability is exercised to perform some task on the resource.

Also, a small nit. "the resource" in the sentence above could be replaced with "a resource". We aren't referring to a specific resource, and "a" is better at implying some resource, doesn't matter which, just that it is associated with the capability.

README.md
```diff
@@ -301,42 +387,65 @@ ownershipscheme = "my:" kind ["@" did]
...
+``` abnf
+delegatescheme = "as:" did ":" kind
+kind = "*" / <scheme>
+```
 
-### 4.1.2 Action
+### 4.2.2 Action
 
-The action for `my:*` MUST be `*`.
+The action for `my:*` or `as:*` MUST be the [superuser action `*`](#41-superuser). Another ability would not be possible, since any other ability cannot be guarinteed to work across all resource types (e.g. it's not possible to `crud/UPDATE` an email address). Recall that the superuser action is special in that it selects the maximum possible action for any resource.
```
bgins, line 403: > cannot be ~~guarinteed~~ guaranteed to work

Typo. Guaranteed, not guarinteed.

README.md
```diff
@@ -301,42 +387,65 @@ ownershipscheme = "my:" kind ["@" did]
...
 ``` json
 {"with": "my:*", "can": "*"}
+{"with": "as:did:key:z6MkiTBz1ymuepAQ4HEHYSF1H8quG5GLVVQR3djdX3mDooWp:*", "can": "*"}
+
+{"with": "my:mailto", "can": "msg/SEND"}
+{"with": "as:did:key:z6MkiTBz1ymuepAQ4HEHYSF1H8quG5GLVVQR3djdX3mDooWp:mailto", "can": "msg/SEND"}
 ```
 
-For `my` capabilities scoped to some scheme, the action MUST be one normally associated with that resource. 
+For `my` and `as` capabilities limited to some scheme, the action MUST be one normally associated with that resource. As it belongs to every action heirarchy, this MAY be the [superuser action `*`](#41-superuser).
```
bgins, line 413: > As it belongs to every action ~~heirarchy~~ hierarchy,

Typo. Hierarchy, not heirarchy.

README.md
```diff
@@ -301,42 +387,65 @@ ownershipscheme = "my:" kind ["@" did]
...
 The `prf` URI scheme defines addressing for UCANs and their fields.
 
 ``` abnf
-prf = "prf" selector
+prf = "prf:" selector
 selector = "*" / 1*DIGIT
 ```
 
-`prf:*` represents all of the UCANs in the current proof scope. The witnesses in the `prf` field may be referenced by their index, starting at 0: `prf:0`
+`prf:*` represents all of the UCANs in the current proofs array. The witnesses for the current UCAN MAY be referenced by their index in the `"prf"` field. If selecting a particular witness (i.e. not the wildcard), then a zero-indexed MUST be used. The first UCAN would be selected by `prf:0`, the second by `prf:1`, and so on. By virtue of the indexing scheme, selections MUST be performed on the current UCAN only, and cannot recursively on nested witnesses.
```
bgins, line 431: > If selecting a particular witness (i.e. not the wildcard), then a zero-indexed MUST be used.

Should "zero-indexed" be followed by something? Like "a zero-indexed scheme"? Or possibly, this could be re-written as

> If selecting a particular witness (i.e. not the wildcard), then ~~a zero-indexed~~ zero-indexing MUST be used.

bgins: > and cannot ~~recursively~~ recurse on nested witnesses.

I think this should be "recurse". Alternatively, this could be something like "and cannot recursively traverse nested witnesses."

README.md
```diff
@@ -301,42 +387,65 @@ ownershipscheme = "my:" kind ["@" did]
...
 
-### 4.2.2 `prf` Actions
+Further selection of capabilities inside fo specific witnesses MUST NOT be a valid parsing of this URI. For instance, `prf:0:mailto` MUST NOT be a valid `prf` URI.
```
bgins, line 433: > Further selection of capabilities inside ~~fo~~ of specific

Typo. Of, not fo.

README.md
```diff
@@ -301,42 +387,65 @@ ownershipscheme = "my:" kind ["@" did]
...
+
+### 4.3.2 `prf` Actions
+
+The `prf` scheme MUST accept the following action: `ucan/DELEGATE`. This redelegates all of the capabilities in the selected witness(es).
```
bgins, line 437: The second sentence could be made stronger by repeating "action". Something like:

> This action redelegates all of the capabilities in the selected witness(es).

README.md
```diff
@@ -348,21 +457,37 @@ If any of the following criteria are not met, the UCAN MUST be considered invali
...
 ## 5.2 Principal Alignment
 
 In delegation, the `aud` field of every witness MUST match the `iss` field of the outer UCAN (the one being delegated to). This alignment MUST form a chain all the way back to the originating principal for each resource.
 
-An agent discharging a capability MUST verify that the outermost `aud` field matches its own DID. If they do not match, the associated action MUST NOT be performed.
+### 5.2.1 Invocation Recipient Validation
+
+An agent discharging a capability MUST verify that the outermost `aud` field _matches its own DID._ If they do not match, the associated action MUST NOT be performed. This is REQUIRED in order to prevent the misuse of UCANs in an unintended context.
+
+The following UCAN fragment would be valid to invoke as `did:key:zH3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV`. Any other agent  MUST NOT accept this UCAN. For example, `did:key:z6MkiTBz1ymuepAQ4HEHYSF1H8quG5GLVVQR3djdX3mDooWp` MUST NOT run the action associated to that capability.
```
bgins, line 472: > Any other agent  MUST NOT

There is an extra space between "agent" and "MUST".

README.md
```diff
@@ -348,21 +457,37 @@ If any of the following criteria are not met, the UCAN MUST be considered invali
...
+
+``` js
+{
+  "aud": "did:key:zH3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV",
+  "iss": "did:key:zAKJP3f7BD6W4iWEQ9jwndVTCBq8ua2Utt8EEjJ6Vxsf",
+  // ...
+}
+```
+
+A good litmus test for invocation validity by a discharging agent is to check if they would  be able to create a valid delegation for that capability.
```
bgins, line 482: > to check if they would  be able to create

There is an extra space between "would" and "be".

README.md
```diff
@@ -450,8 +583,28 @@ Thank you to [Brendan O'Brien](https://github.com/b5) for real-world feedback, t
...
 
 Thanks to the entire [SPKI WG](https://datatracker.ietf.org/wg/spki/about/) for their closely related pioneering work.
 
 We want to especially recognize [Mark Miller](https://github.com/erights) for his numerous contributions to the field of distributed auth, programming languages, and computer security writ large.
+
+# 9. FAQ
+
+## 9.1 What prevents an unauthorized party from using an intercepted UCAN?
+
+UCANs always contain the information about the sender and receiver. It is signed by the sender (the `iss` field DID), and so can only be created by an agent in posession of the relevant private key. The recipient (the `aud` field DID) is required to check that the field matches their DID. These two checks taken together secures against use by an unauthorized party.
```
bgins, line 598: The second sentence can be strengthened a bit by repeating UCAN. Something like:

> A UCAN is signed by the sender (the `iss` field DID)

Typo in the last sentence:

> These two checks taken together ~~secures~~ secure against use by an unauthorized party.

README.md
```diff
@@ -208,16 +283,16 @@ The OPTIONAL `fct` field contains arbitrary facts and proofs of knowledge. The e
...
-The attenuation scope (i.e. UCAN output, or "caveats") MUST be an array of heterogeneous access scopes (defined below). This array MAY be empty.
+The attenuations (i.e. UCAN output, or "caveats") MUST be an array of heterogeneous capabilities (defined below). This array MAY be empty.
 
 This array MUST contain some or none of the following:
-1. A strict subset (attenuation) of the witnesses
-2. Witnesses originated by the `iss` DID (i.e. by parenthood)
-3. Witnesses composed from other witnesses (see rights amplification)
+1. A strict subset (attenuation) of the capability scope from the `prf` field
+2. Capabilities composed from multiple witnesses (see rights amplification)
+3. Capabilities originated by the `iss` DID (i.e. by parenthood)
```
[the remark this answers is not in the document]
expede: Parenthood is the technical term for "resource created by this agent". Ambient authority is a broader concept that includes anything that is not captured by the capability (often identity based).

README.md
expede: Good catch!

README.md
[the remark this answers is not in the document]
bgins: Thanks for the clarification, I was unaware of the term parenthood. It definitely makes more sense than ambient authority in this context. 👍

PR #26, opened 2022-02-17
Move Irakli to authors section

expede:
@Gozala per my suggestion earlier

Rationale (for posterity): Irakli has been a great collaborator, and has suggested several nontrivial improvements to he spec, from wording (e.g. `can`) to pushing for a more IPLD friendly format. Not all of that has landed yet (though some has), but probably more of it will in `0.9`. It only seems right that he get the credit on 0.8, and 0.9 for sure!


Gozala (approved):
Thanks, honored to be on that list

PR #27, opened 2022-02-18
Minor edits

mistermoe:
Hey there!

While reading, I ran into what i think may be two typos, so I figured i'd submit a PR to fix them. Feel free to close the PR if they're not actually typos

Also, many props for both the concept and how the spec is written. Both are incredibly intuitive and easy to grok. 😃


expede (approved):
haha whoops! Thank you @mistermoe 🎉

PR #28, opened 2022-02-24
Fix `ownershipscheme` ABNF grammar

matheus23:
According to @expede:

> Yeah it’s a relic from an old ABNF

Where "it" refers to tagging the DID in the `my:*` scheme.


expede (approved):
Thanks @matheus23!

PR #34, opened 2022-03-20
Update per Grammarly

expede:
Per @bgins, I ran the spec through Grammarly. Not everything made sense for spec style, but I took many of the suggestions 👍 

There's still a lot of passive voice, but that's going to take a lot more work. It is also sometimes to rephrase when respecting RFC 2119.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
 
 # 0. Abstract
 
-User-Controlled Authorization Network (UCAN) is a trustless, secure, local-first, user-originated authorization and revocation scheme. It provides public-key verifiable, delegable, expressive, openly extensible [capabilities](https://en.wikipedia.org/wiki/Object-capability_model) by extending the familiar [JWT](https://datatracker.ietf.org/doc/html/rfc7519) structure. UCANs achieve public verifiability with chained certificates and [decentralized identifiers (DIDs)](https://www.w3.org/TR/did-core/). Verifiable chain compression is enabled via [content addressing](https://en.wikipedia.org/wiki/Content-addressable_storage). Being encoded with the familiar JWT, UCAN improves the familiarity and adaptability of schemes like [SPKI/SDSI](https://theworld.com/~cme/html/spki.html) for web and native application contexts. UCAN allows for the creation and discharge of authority by any agent with a DID, including traditional systems and peer-to-peer architectures beyond traditional cloud computing.
+User-Controlled Authorization Network (UCAN) is a trustless, secure, local-first, user-originated authorization and revocation scheme. It provides public-key verifiable, delegable, expressive, openly extensible [capabilities](https://en.wikipedia.org/wiki/Object-capability_model) by extending the familiar [JWT](https://datatracker.ietf.org/doc/html/rfc7519) structure. UCANs achieve public verifiability with chained certificates and [decentralized identifiers (DIDs)](https://www.w3.org/TR/did-core/). Verifiable chain compression is enabled via [content addressing](https://en.wikipedia.org/wiki/Content-addressable_storage). Being encoded with the familiar JWT, UCAN improves the familiarity and adoptability of schemes like [SPKI/SDSI](https://theworld.com/~cme/html/spki.html) for web and native application contexts. UCAN allows for the creation and discharge of authority by any agent with a DID, including traditional systems and peer-to-peer architectures beyond traditional cloud computing.
 
 ## Language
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,21 +89,21 @@ Unlike many authorization systems where a service controls access to resources i
...
 
 ## 2.2 Action
 
-An action is performed against some resource. Each action MAY have its own semantics. They MAY be unary, support a hierarchy, be monotone, partial orders, and so on. Actions MAY be general and applicable to many kinds of resource, or tied to a specific one.
+An action MUST be performed against some resource. Each action MAY have its own semantics. For example, actions MAY be unary, support a hierarchy, be monotone, form partial orders, etc. In addition, actions MAY be general and applicable to many kinds of resources or tied to a specific one.
 
-For instance, `wnfs/APPEND` is an action for WebNative filesystem paths. The action `wnfs/OVERWRITE` also implies the capacity to append. Email has no such tiered relationship. One can `email/SEND`, but there is no concept of a ”super send”.
+For instance, `wnfs/APPEND` is an action for WebNative filesystem paths. The action `wnfs/OVERWRITE` also implies the capacity to append. An email has no such tiered relationship. One can `email/SEND`, but there is no concept of a" super send."
```
bgins, line 98: > but there is no concept of ~a" super send."~ a "super send."

Add a space after "a" and remove one from before "super".

README.md
```diff
@@ -350,9 +350,9 @@ The only reserved ability MUST be the unnamespaced [`"*"` or "superuser"](#41-su
 
 ### 3.2.5 Proof of Delegation
 
-The `prf` field MUST contain UCAN witnesses (the ”inputs” of a UCAN). As they need to be independently verifiable, proofs MUST either be the fully encoded version of a UCAN (including the signature), or the content address of the relevant proof. Attenuations not covered by a proof in the `prf` array MUST be treated as owned by the issuer DID.
+The `prf` field MUST contain UCAN witnesses (the" inputs" of a UCAN). As they need to be independently verifiable, proofs MUST either be the fully encoded version of a UCAN (including the signature) or the content address of the relevant proof. Attenuations not covered by a proof in the `prf` array MUST be treated as owned by the issuer DID.
```
bgins, line 353: > ~(the" inputs" of a UCAN)~ (the "inputs" of a UCAN)

Move space outside of opening quote.

README.md
```diff
@@ -526,35 +526,35 @@ A revocation message MUST conform to the following format:
 }
 ```
 
-This format makes it easy to select the relevant UCAN, confirm that the issuer is in the proof chain for some or all capabilities, and validate that the revocation itself was signed by the same issuer. 
+This format makes it easy to select the relevant UCAN, confirm that the issuer is in the proof chain for some or all capabilities, and validate that the revocation was signed by the same issuer.
 
 Any other witnesses in the selected UCAN not issued by the same DID as the revocation issuer MUST be treated as valid.
 
-Revocations MAY be deleted once the UCAN they reference expires or otherwise becomes invalid via its proactive mechanisms.
+Revocations MAY be deleted once the UCAN the reference expires or otherwise becomes invalid via its proactive mechanisms.
```
bgins, line 533: > the UCAN ~the~ they reference

"they" looks like the right word here.

bgins (approved):
cat-editor

Looks great! It's nice to have Grammarly to catch all of these little details. Plus it often gives great conciseness and word choice recommendations.

I caught a couple of places where Grammarly had issues with markdown syntax and spaces, plus a couple of instances where I wasn't sure about its word choices. In general, this is a nice overall improvement!

README.md
expede: Woah weird. I wonder if there's a mismatched quote somewhere? Looks like that happened as part of the conversion to consistent quotes.

Good catch!

README.md
```diff
@@ -132,11 +132,11 @@ Merging capability scopes MUST follow set semantics, where the result includes a
                  scope
 ```
 
-The capability scope is the total rights of the authorization space down to the relevant volume of authorizations. Individual capabilities MAY overlap; the scope is the union. With the exception of [rights amplification](#54-rights-amplification), every individual delegation MUST have equal or narrower capabilties from their delegator. Inside this content space, you can draw a boundary around some resource(s) (their type, identifiers, and paths or children), and their capabilities.
+The capability scope is the total rights of the authorization space down to the relevant volume of authorizations. Individual capabilities MAY overlap; the scope is the union. Except for [rights amplification](#54-rights-amplification), every unique delegation MUST have equal or narrower capabilities from their delegator. Inside this content space, you can draw a boundary around some resource(s) (their type, identifiers, and paths or children) and their capabilities.
 
 For example, given the following scopes against a WebNative filesystem, they can be merged as follows:
 
-```js
+"`js
```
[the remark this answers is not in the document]
expede: 😆

expede:
Wow, Grammarly DOES NOT like Markdown and RFC 2119

PR #37, opened 2022-03-24
Add UCAN v0.8 Test Fixtures

### LINKED ISSUE #7: v0.8 test fixtures

appcypher:
Fixes #7 

PR adds test fixtures for complying with UCAN v0.8.
While working on this [I used a utility](https://github.com/fission-suite/ucan-fixture-gen/blob/main/src/samples.ts) other people might find useful when working with UCAN fixtures in the future.
https://github.com/fission-suite/ucan-fixture-gen

Reviews welcome.


bgins:
Hey @appcypher. Thanks for adding these test fixtures! They are looking great and are much appreciated. 🙏 

One detail that should be changed is the `ucv` in these fixtures. The `ucv` should [use semantic versioning](https://github.com/ucan-wg/spec#31-header). Instead of `0.8`, could we change these to `0.8.1` (or `0.8.0`, if they aren't meant to cover the latest spec version.)

appcypher:
@bgins @expede Forgot to mention that the versions have been fixed

tests/0.8.1/valid.json
```diff
@@ -0,0 +1,376 @@
...
+            "from": "example.com"
+          }
+        ],
+        "att": [],
+        "prf": []
+      }
+    }
+  },
+  {
+    "comment": "Delegated UCAN can have multiple valid proofs",
```
bgins, line 297: What would you think about rephrasing this comment to "Delegated UCAN is valid with multiple valid proofs"? This would strengthen the idea that a UCAN can have multiple proofs _and_ is valid because of the capabilities granted by each of those proofs.

tests/0.8.1/valid.json
```diff
@@ -0,0 +1,376 @@
...
+        ],
+        "prf": [
+          "eyJhbGciOiJFZERTQSIsInR5cCI6IkpXVCIsInVjdiI6IjAuOC4xIn0.eyJpc3MiOiJkaWQ6a2V5Ono2TWt3YTUydEh5UHMzY1J5NlNja1hEd2R4ZEI2UjRWWlhwemNQbUg2cjV1MjRqUSIsImF1ZCI6ImRpZDprZXk6ejZNa3BXUEVmdzJ5VVh0bU16NlNXNGh5ZTdYcGpCUVNOMUVXamU2dDEybVJEQzdWIiwiZXhwIjo0ODAzODkwNTQ0LCJhdHQiOlt7IndpdGgiOiJkYjovL3RhbWVkdW4uZmlzc2lvbi5hcHAvdXNlcnMiLCJjYW4iOiJkYi9SRUFEIn1dLCJwcmYiOltdfQ.f422tsTyFwa6ULsBeDHHLxufpty1r79EdTkMSC0znwypiVSYJi5GbMChO-oRbVyBMXX50QBkm1YXGSoo_qOgBg",
+          "eyJhbGciOiJFZERTQSIsInR5cCI6IkpXVCIsInVjdiI6IjAuOC4xIn0.eyJpc3MiOiJkaWQ6a2V5Ono2TWtyQ0dObVdVOW9oR2d5WGt6YTRxNlFpYXB1WlF3NkpndU1xV1M3U1JMTlZzWCIsImF1ZCI6ImRpZDprZXk6ejZNa3BXUEVmdzJ5VVh0bU16NlNXNGh5ZTdYcGpCUVNOMUVXamU2dDEybVJEQzdWIiwiZXhwIjo0ODAzODkwNTQ0LCJhdHQiOlt7IndpdGgiOiJkYjovL3RhbWVkdW4uZmlzc2lvbi5hcHAvdXNlcnMiLCJjYW4iOiJkYi9XUklURSJ9XSwicHJmIjpbXX0.NOuYmki3fQWQ09x4XseAWooCoSB6UZwqoZJ0KURUnN9uvgvRLI-FwNAv0y2c5JvvzrE7CxJMK_VvePTAH_y8DQ"
+        ]
+      }
+    }
+  },
+  {
+    "comment": "UCAN attenuation is valid syntax",
```
bgins, line 327: > "UCAN attenuation ~is~has valid syntax"

Typo. Has, not is.

tests/0.8.1/valid.json
```diff
@@ -0,0 +1,376 @@
...
+        "iss": "did:key:z6MkpWPEfw2yUXtmMz6SW4hye7XpjBQSN1EWje6t12mRDC7V",
+        "aud": "did:key:z6MkpkQPXHuxoqH7oukdohXbQ8uU8QE8KpAiuYpCEhwBvhff",
+        "exp": 4803890544,
+        "att": [],
+        "prf": []
+      }
+    }
+  },
+  {
+    "comment": "Payload `nbf` is valid",
```
bgins, line 236: Is this test redundant with "UCAN is ready to be used"? If we are testing the same thing, we should only keep one.

tests/0.8.1/valid.json
```diff
@@ -0,0 +1,376 @@
...
+        "aud": "did:key:z6MkpkQPXHuxoqH7oukdohXbQ8uU8QE8KpAiuYpCEhwBvhff",
+        "nbf": 1648130544,
+        "exp": 4803890544,
+        "att": [],
+        "prf": []
+      }
+    }
+  },
+  {
+    "comment": "Payload `exp` is valid",
```
bgins, line 255: Might be testing the same thing as "UCAN has not expired".

tests/0.8.1/invalid.json
```diff
@@ -0,0 +1,842 @@
...
+    "comment": "UCAN sections contain invalid base64 characters",
+    "token": "@@JhbGciOiJFZERTQSIsInR5cCI6IkpXVCIsInVjdiI6IjAuOCJ9.eyJpc3MiOiJkaWQ6a2V5Ono2TWtySnpWZVk3RXd2cVdNc3ROeDJ4SDdiZGVnYWREZGFjOHVGams2NTlnemJkeCIsImF1ZCI6ImRpZDprZXk6ek0wMG04RHhXU3dRaGhaWWJnUGprQ05qbUx2dmEzRDdxQnNHUHZ3ejJneW5TaWFKIiwiZXhwIjoxNjQ4MDc2MDQ0LCJhdHQiOltdLCJwcmYiOltdfQ.vLgf-O3Xrc94toabVJKoI15EmAO-d9SqHGFbxTkw3wyEZb-Q7YXA-6mWJv6b-b7kFQHiilMcdIl5s5IuXxsSAw",
+    "assertions": {
+      "validationErrors": [
+        "base64Invalid"
+      ]
+    }
+  },
+  {
+    "comment": "UCAN header section is missing or invalid",
```
bgins, line 12: We might want to say "UCAN header section is missing" and add a separate test case for an invalid header.

tests/0.8.1/invalid.json
```diff
@@ -0,0 +1,842 @@
...
+        "att": [],
+        "prf": []
+      },
+      "validationErrors": [
+        "headerMissingOrInvalid"
+      ]
+    }
+  },
+  {
+    "comment": "UCAN payload section is missing or invalid",
```
bgins, line 28: Separate cases for missing and invalid might be good here as well.

tests/0.8.1/invalid.json
```diff
@@ -0,0 +1,842 @@
...
+        "typ": "JWT",
+        "ucv": "0.8.1"
+      },
+      "validationErrors": [
+        "payloadMissingOrInvalid"
+      ]
+    }
+  },
+  {
+    "comment": "UCAN signature is missing or invalid",
```
bgins, line 42: Separate cases for missing and invalid might be good here too.

tests/0.8.1/invalid.json
```diff
@@ -0,0 +1,842 @@
...
+        "att": [],
+        "prf": []
+      },
+      "validationErrors": [
+        "nbfNotReady"
+      ]
+    }
+  },
+  {
+    "comment": "Witnesses expire before the delegated UCAN",
```
bgins, line 106: The expiration times for the delegate and witness are `5435042552`. Did you mean edit the witness forward in time for this test case?

tests/0.8.1/invalid.json
```diff
@@ -0,0 +1,842 @@
...
+          "eyJhbGciOiJFZERTQSIsInR5cCI6IkpXVCIsInVjdiI6IjAuOC4xIn0.eyJpc3MiOiJkaWQ6a2V5Ono2TWtyaHc2cUtDRlBEVlUxemd1MU5yNHI2WkxTcDJkSjhTUlFkZTZMaVdNS0NkRCIsImF1ZCI6ImRpZDprZXk6ejZNa2dhQWFWU3pFOVhxYW5yYkZEMUo5UkdURmNXVTdOOUJQZ2pTV20xN3BrMW1CIiwibmJmIjo0ODAzODkwNTUyLCJleHAiOjU0MzUwNDI1NTIsImF0dCI6W10sInByZiI6W119.-s-lC0nJMZk0zlnjSalo4IkQ_fslzhyu464GqQOhIyo7n0B5uvZTaFOIZ8Zsaled8yDAKe93btXwljXJiEgiAA"
+        ]
+      },
+      "validationErrors": [
+        "expWitnessTimeBoundExceeded"
+      ]
+    }
+  },
+  {
+    "comment": "Witness issuer audience did does not align with delegated issuer did",
```
bgins, line 153: It may be easier to distinguish the decentralized identifiers in this comment if we change "did" to "DID".

tests/0.8.1/invalid.json
```diff
@@ -0,0 +1,842 @@
...
+          "eyJhbGciOiJFZERTQSIsInR5cCI6IkpXVCIsInVjdiI6IjAuNyJ9.eyJpc3MiOiJkaWQ6a2V5Ono2TWtwemREYk1UQzRXcUtMMTdEd1BqYjQyU2JLWlJkYVpOaUdSWEdtTThGbmtkTSIsImF1ZCI6ImRpZDprZXk6ejZNa2dhQWFWU3pFOVhxYW5yYkZEMUo5UkdURmNXVTdOOUJQZ2pTV20xN3BrMW1CIiwiZXhwIjo0ODAzODkwNTUyLCJhdHQiOltdLCJwcmYiOltdfQ.PlPmieF7q0E-NrduHSFInDuWAXpmtuS0puzp-jnWVdCRxFzasizJTO2Z8-lmWB7GC9mv11_ttsJUDZuZGaJYAA"
+        ]
+      },
+      "validationErrors": [
+        "prfWitnessVersionMismatch"
+      ]
+    }
+  },
+  {
+    "comment": "Witness does not exist",
```
bgins, line 199: This comment could be made more specific by mentioning proof schemes. One possibility would be "Witness referenced in prf scheme does not exist".

tests/0.8.1/invalid.json
bgins: Got further down into the invalid cases, and I noticed there are cases of headers with fields missing. That's what I would consider and "invalid" header. No need to add more test cases (same goes for payload comment below, but we could use a garbage signature test case.)

tests/0.8.1/invalid.json
```diff
@@ -0,0 +1,842 @@
...
+        "att": [],
+        "prf": []
+      },
+      "typeErrors": [
+        "nbfWrongType"
+      ]
+    }
+  },
+  {
+    "comment": "Payload `exp` field should be a did:key string",
```
bgins, line 601: > "Payload `exp` field should be a ~did:key string~ number"

The test is right, but the comment is incorrect.

tests/0.8.1/invalid.json
```diff
@@ -0,0 +1,842 @@
...
+        "att": [],
+        "prf": []
+      },
+      "typeErrors": [
+        "fctWrongType"
+      ]
+    }
+  },
+  {
+    "comment": "Payload `prf` field should be an array of string",
```
bgins, line 686: > "Payload `prf` field should be an array of ~string~strings"

Typo. Strings, not string.

tests/0.8.1/invalid.json
```diff
@@ -0,0 +1,842 @@
...
+        "att": [],
+        "prf": 1
+      },
+      "typeErrors": [
+        "prfWrongType"
+      ]
+    }
+  },
+  {
+    "comment": "Payload `prf` field should be an array of string",
```
bgins, line 707: > "Payload prf field should be an array of stringstrings"

Typo. Strings, not string.

bgins (approved):
Fantastic! Thrilled to have these tests, great work on them! 💯 🎉 

I left a few comments, mostly nitpicky stuff around the comments. Thanks again for writing these! 🙏

tests/0.8.1/invalid.json
appcypher: The reason for having it as `missing or invalid` here is because it can be non-trivial to detect the intention of the UCAN author. For example, the header may in fact not be missing but malformed maybe by having some additional non-compliant fields or missing some compliant fields. If it just says `missing` then we are forcing the implementer to detect all of those cases.
I can change the wording to `malformed`.

appcypher:
Fixed @bgins

tests/0.8.1/invalid.json
bgins: Good point. Reading through the example, it was trivial to see that the header was missing, but testing for it would be challenging. Renaming to `malformed` sounds good to me. 👍

expede (approved):
LGTM 👍 

(@appcypher merge #49 first, and the cla-bot will also sign off)

expede:
@appcypher `ucan-fixture-generator` is fabulous 🙌 Any concerns against us moving it from `fission-suite` to `ucan-wg`?

appcypher:
@expede None and done

PR #58, opened 2022-04-09
Add gozala as CLA signatory

Gozala:


matheus23 (approved):

matheus23:
Thanks @Gozala :pray: 

We need these commits to be signed though, and it looks like that's not the case. Can you [configure commit signing](https://docs.github.com/en/github-ae@latest/authentication/managing-commit-signature-verification/signing-commits) and force-push again?

Gozala:
> Thanks @Gozala 🙏
> 
> We need these commits to be signed though, and it looks like that's not the case. Can you [configure commit signing](https://docs.github.com/en/github-ae@latest/authentication/managing-commit-signature-verification/signing-commits) and force-push again?

Ah I just edited them from the github UI, which is probably whey they didn't got signed. Will do one from my device

Gozala:
@matheus23 Just force pushed signed commit

matheus23 (approved):
Great, thanks!

Issue #68, opened 2022-05-13
spec incompatible test fixture


Gozala:
This fixture seems to be incompatible with a spec

https://github.com/ucan-wg/spec/blob/76e91bd9da67cf796f04760372fe1894e58b1480/fixtures/0.8.1/valid.json#L207-L208

As I understand it it should be `prf:0` instead per
https://github.com/ucan-wg/spec#prf-scheme

expede:
Indeed. Will fix; thanks!

PR #69, opened 2022-05-16
Fix 0.8 fixture for `prf` capability

expede:


expede:
Fixes https://github.com/ucan-wg/spec/issues/68

fixtures/0.8.1/valid.json
```diff
@@ -204,7 +204,7 @@
         "exp": 4804143412,
         "att": [
           {
-            "with": "prf/0",
+            "with": "prf:0",
```
Gozala, line 207: Would not this affect the actual token ? Or was this just a typo here ?

Gozala: As far as I can tell `token` field is also needs to be updated a per https://github.com/ipld/js-dag-ucan/runs/6458958822?check_suite_focus=true#step:5:107

Gozala: Pasting token into https://ucan.xyz/validator also confirmed my suspicion
image

matheus23: Yeah, you're right. I don't know how to recreate the signature correctly with that issuer's key, so here's a different token that should do the same thing:
```
eyJhbGciOiJFZERTQSIsInR5cCI6IkpXVCIsInVjdiI6IjAuOC4xIn0.eyJhdWQiOiJkaWQ6a2V5Ono2TWtnWDVqalJVYnR5c2dnRTRyYUNhcUNYODhBelN2WXE4MVdKa0JvQTFvdDhhZSIsImF0dCI6W3sid2l0aCI6InByZjowIiwiY2FuIjoidWNhbi9ERUxFR0FURSJ9XSwiZXhwIjo0ODA0MTQzNDEyLCJmY3QiOltdLCJpc3MiOiJkaWQ6a2V5Ono2TWtoejNhdENZdGk5QktWUEc4alcyYlFCbnR5RW5ZS3BpTVJreTJuOGg0eVY1VSIsInByZiI6W119.TTqaclWZgPWDIjGk4tdHPHhzUE2NOLjGh7Wk3SX4GCQy9Kw3niD-9Q4ubTv2BoCk_anNEmCrmWUCijBd137pBg
```

I'll edit the PR.

matheus23: Actually, scrap that. The token I pasted above is incorrect, and I now see where the tokens are generated: https://github.com/ucan-wg/ucan-fixture-gen/blob/main/src/samples.ts

matheus23:
Unfortunately, it's impossible to regain access to the original private key behind the fixtures, so the diff is larger than neccessary. I've used the ucan-fixture-gen project to re-generate these fixtures correctly (and I've opened a PR over there https://github.com/ucan-wg/ucan-fixture-gen/pull/1).

There was another fixture using an incorrect delimiter, I've fixed that one, too. :v:

Should be good now.

PR #55, opened 2022-04-06
UCAN v0.9.0

expede:
* 👉 [Preview Render 📃](https://github.com/ucan-wg/spec/blob/0.9/README.md)
* #35 

# Change Log 🪙🪵

## Added 📥

* [x] Explicitly restrict keys to RSA, ECDSA, and EdDSA per @ukstv
* [x] Standardize on "proof" instead of "witness" (see https://github.com/ucan-wg/spec/discussions/23)
  * At minimum it leaves the door open for renaming later; we already call them proofs
* ~~[ ] Revocation spec~~ Staying the same, see comment: https://github.com/ucan-wg/spec/pull/55#discussion_r844809873
* [x] Standardize on "ability" and not "action"
* [x] Break out caveats fields, per @gozala https://github.com/ucan-wg/spec/discussions/61
* [x] Replace `my:...` URI with `"my"` parenthood field
* [x] Changes to UCAN DAG representation
  * [x] ~~CARv2~~ Agnostic collections
  * [x] Proofs enforced as CIDs _only_

## Removed 📤

* [x] Remove IPFS/DHT requirement
* [x] Remove URI selection by scheme (e.g. no more `"with": "dns:*"`)
  * This was actually very bad and broke the URI contract

## Move Out of Core Spec 🪁

* ~~[x] Consider extending `my:` to all parenthood cases, even specific values (CC @matheus23)~~
* Transport agnostic point-to-point protocol
* HTTP cache header protocol <- HTTP stuff should live in own spec
* Resource selection (e.g. `"with": "ucan:uri:dns"`)

# Resolves 🛠️

* Closes #61
* Closes #77

PR #87, opened 2022-07-25
0.9: Namespace all UCAN-spec required resources (e.g. my and as)

expede:
✨ **NB: this is a PR against #55, not `main`** ✨

A suggestion from Irakli and Hugo a few weeks back was to namespace all of the "first party" resources like `ucan:`, `my:`, and `as:`. Of course this creates a conflict with the existing `ucan:` URI.

This uses more characters, but has a few advantages:

* Leaves the top-level namespace more clean (someone else can claim `my:` later)
* Clear that these are all UCAN-specific ("special") resources

I broke this into its own PR because while I'm 99% sure this is a good idea, I want to play with the idea and exact labels a bit (such as `ucan:token:...`).

An open question is if the `can: *` super user ability should also be prefixed for consistency as `can: ucan/*`. I don't think that this actually adds clarity, but can see arguments both ways. FWIW, we captured a [this short HackMD](https://hackmd.io/jO5xcBDPQp6WZINuX8FJYA) as part of the live call, and it doesn't have the `ucan/` prefix.

@matheus23 any thoughts?

# Change Log

| Before | After |
|--------|-------|
| `my:dns` | `ucan:my:dns` |
| `as:did:key:zAlice` | `ucan:as:did:key:zAlice` |
| `ucan:bafyabcdef` | `ucan:token:bafyabcdef` |

### DIFF
--- a/.custom-words.txt
+++ b/.custom-words.txt
@@ -99,6 +99,7 @@ subdelegate
 tradeoffs
 trustless
 trustlessly
+ucan
 un
 unary
 unresolvable
--- a/README.md
+++ b/README.md
@@ -344,8 +344,8 @@ The same resource MAY be addressed with several URI formats. For instance, a dat
 | URI                                            | Meaning                                                         |
 | ---------------------------------------------- | --------------------------------------------------------------- |
 | `{"with": "mailto:username@example.com", ...}` | A single email address                                          |
-| `{"with": "my:*", ...}`                        | All resources that the iss has access to, including future ones |
-| `{"with": "my:dnslink", ...}`                  | All DNSLinks that the iss has access to, including future ones  |
+| `{"with": "ucan:my:*", ...}`                   | All resources that the iss has access to, including future ones |
+| `{"with": "ucan:my:dnslink", ...}`             | All DNSLinks that the iss has access to, including future ones  |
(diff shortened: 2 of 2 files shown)


matheus23 (approved):
> I'm 99% sure this is a good idea

Hm, I'm not super convinced. We're already in a "ucan namespace", and I don't see value in someone claiming `my` for themselves.

But I'm also not opposed to this, so this can go ahead :+1: 

> An open question is if the `can: *` super user ability should also be prefixed for consistency as `can: ucan/*`.

Honestly, I can see `can: "ucan/*"` vs. `can: "*"` going either way.
- `ucan/*` is more consistent with `ucan/DELEGATE` but
- requiring `ucan/*` only gets us the ability to define `*` as an ability for other purposes. I don't think that's super useful?

expede:
> We're already in a "ucan namespace", and I don't see value in someone claiming my for themselves.

The argument really is to keep it cleaner. I should perhaps have been crisper on the use of namespace: "URI namespace".

> Honestly, I can see can: "ucan/*" vs. can: "*" going either way.

Same

> requiring ucan/* only gets us the ability to define * as an ability for other purposes. I don't think that's super useful?

You can do that either way (`mycoolapp/*`). I'm fine with `*` being "special" in the global namespace; I think it's clear what it does.

PR #88, opened 2022-07-27
Break out `my:` field & delegation selectors

expede:


README.md
```diff
@@ -431,108 +480,38 @@ Which in a JSON representation would resolve to the following table:
...
-
-### 4.2.2 `ucan:my` Wildcard
-
-The wildcard `ucan:my:*` resource MUST be taken to mean "everything" (all resources of all types) that are owned by the current DID.
-
-### 4.2.3 Subschemes
-
-A "sub-scheme" MAY be used to delegate some of that scheme controlled by parenthood. For example, `my:dns` delegates access to all DNS records. `my:mailto` selects all owned email addresses controlled by this user.
-
-### 4.2.4 `ucan:as:` Redelegation
```
expede, line 492: _Controversial,_ but @matheus23 we can move this into an extension spec for now. It's very important to our use case at Fission, but may not be terribly important to others.

I'm also open to putting this back in, but it's safe to pull out here in a subbranch.

README.md
```diff
@@ -314,7 +315,60 @@ The OPTIONAL `fct` field contains arbitrary facts and proofs of knowledge. The e
...
+```js
+{
+  ...,
+  "my": [
+    {
+      "with": "mailto:alice@example.com",
+      "can": "msg/send"
+    },
+    {
+      "with": "dns:*", // FIXME breaks the URI
```
expede, line 345: We can't actually do `<uri>:*` with impunity. Everything that comes after the URI scheme is defined by that scheme, and thus `*` may have a special meaning. We got away with this before by prefixing with `ucan:`, which changes the rules because it's nested.

Thinking out loud, we could use a `ucan:all:` prefix or something? I'll think about it a bit more.

expede: 🤔 `ucan:*:dns`?

PR #104, opened 2022-09-08
Roles Table

expede:
[:scroll: Preview](https://github.com/ucan-wg/spec/blob/roles/README.md#21-roles)


expede:
This proposal is not sacred and naming is always hard. Please do make suggestions / changes as needed :)

expede:
Delegator/delegate may be too close visually. We could also use Issuer / Audience.

expede:
Testing webhook

bgins:
> Delegator/delegate may be too close visually. We could also use Issuer / Audience.

Agreed that delegator and delegate a bit close. To me, these terms more strongly indicate passing capabilities than issuer and audience. Do we want to differentiate on whether capabilities are granted or not? 

One case to consider here is establishing an AWAKE session, where a UCAN indicates capability without delegating anything (https://github.com/ucan-wg/awake#33-responder-establishes-point-to-point-session)

> In this step, the Responder MUST prove that they have access to the requested resources. This is used to establish trust in the capabilities of the Responder, but MUST NOT actually delegate anything. 

In Discord chat, I suggested `Authority` and `Delegate` as an alternative, but "authority" is an overloaded term and this probably wouldn't work out. Consider the following for example

> An authority with root authority grants a subset of capabilities to a delegate.

Not great! 🙅 

One other dimension we might consider -- how de we name an agent to evaluates a UCAN before actually using a resource? I'm not sure if Validator fully covers this. 

As a practical example, we recently added an endpoint for uploading files to a Fission app. A user registers their DID with the Fission server and delegates to the Fission server when they upload a file. The user can also delegate to another agent to upload on their behalf. Following that, the other agent delegates to the Fission server when they upload a file for the user. 

The other agent can check the UCAN is valid and that it grants sufficient capability, but it doesn't know for sure that the Fission app resource exists. Instead, it trusts the UCAN and verification is delayed until it attempts to upload a file.

In this scenario, is the other agent a Validator? Is the server the only Validator? The difference between the two is whether an action is taken with a resource on checking the UCAN.

README.md
```diff
@@ -87,7 +87,19 @@ Unlike many authorization systems where a service controls access to resources i
...
+
+There are several roles that an agent MAY assume:
+
+| Name      | Description | 
+| --------- | ----------- |
+| Agent     | The general class of entities and principals that interact with a UCAN |
+| Principal | A DID listed in a UCAN's `iss` or `aud` field |
+| Delegator | The principal signing the current UCAN, and listed in the `iss` field |
+| Delegate  | The principal being delegated to in the current UCAN, listed in the `aud` field |
+| Validator | Any party that interprets a UCAN to determine that it is valid, and which capabilities it grants |
```
hugomrdias, line 100: The way i think about it is:
- `Agent` can be an Issuer or a Audience
- `Principal` generic denomination for DID
- `Issuer` is a principal with a private key
- `Audience` is just a Principal that we need a special word for

hugomrdias: Actually `Agent` and `Issuer` can be the same thing depends on the side.
Example:

From Agent1 side he is the Agent (Principal with private key) and Audience is just a DID (Agent2 DID).
From Agent2 side reverse.

So Agent can't really be Issuer or Audience, its more a question of locality (local vs remote).

So maybe this is enough:
- `Principal` is just a DID (may or may not have a keypair)
- `Agent` is a local `Principal` that can issue UCANs with a keypair to another `Principal`

Gozala: @hugomrdias I think thinking of `Issuer` as an entity with private key is misleading, because it depends on the perspective. On the sender side UCAN issuer is the entity with private key as it needs to sign that UCAN, however on the receiver side UCAN issuer is no longer entity with a private key or at least representation of it can not sign anything.

I am afraid `Audience` is the same you're on the receiver end audience will be an entity with signing capabilities, however on the sender side audience will be a role with verification capabilities.

README.md
```diff
@@ -87,7 +87,19 @@ Unlike many authorization systems where a service controls access to resources i
...
 
-## 2.1 Resource
+## 2.1 Roles
+
+There are several roles that an agent MAY assume:
+
+| Name      | Description | 
+| --------- | ----------- |
+| Agent     | The general class of entities and principals that interact with a UCAN |
+| Principal | A DID listed in a UCAN's `iss` or `aud` field |
```
Gozala, line 97: How about this instead ?

```suggestion
| Principal | An entity identified by DID, listed in a UCAN's `iss` or `aud` field |
```

Gozala (commented):
Thanks @expede for adding these, it does really help. It however does not necessarily address naming dilemma we're having and I think it would be amazing if we try to tackle that as well. Specifically

`Delegator` on the UCAN issuer side is an entity capable of signing payloads, but on the side of audience it's no longer entity capable of signing payloads. In fact what used to be audience on the issuer side becomes `Delegator` capable of signing payloads.

I'm not sure if these are way to reconcile these, but it would be amazing to have different terms of the principal that can issue vs principal that has issued / delegated.

README.md
```diff
@@ -87,7 +87,60 @@ Unlike many authorization systems where a service controls access to resources i
...
+│   │  │  │  └──────────────┘  │  │  │                      │  │  │                      │   │
+│   │  │  │                    │  │  │                      │  │  │                      │   │
+│   │  │  └────────────────────┘  │  │                      │  │  │                      │   │
+│   │  │                          │  │                      │  │  │                      │   │
+│   │  └──────────────────────────┘  └──────────────────────┘  │  │                      │   │
+│   │                                                          │  │                      │   │
+│   └──────────────────────────────────────────────────────────┘  └──────────────────────┘   │
+│                                                                                            │
+└────────────────────────────────────────────────────────────────────────────────────────────┘
+```
```
expede, line 141: @hugomrdias & @Gozala I based this on a combination of what's in the Discord and comments on this PR. Does this more closely match your picture?

expede: Preview: https://github.com/ucan-wg/spec/blob/roles/README.md#21-roles

bgins: This looks good to me! Owner covers the scenario I described in my comment about. 👌

bgins (approved):
Looks good! The diagram is especially nice for visualizing how the roles are related. 💯

Gozala (approved):
Looks good to me

README.md
hugomrdias: @expede this looks awesome really helpful

PR #116, opened 2022-10-20
fix: broken rights-amplification link

### LINKED ISSUE #115: fix: broken rights amplification URL
The hyperlink to the rights amplification chapter is broken here:
https://github.com/ucan-wg/spec/blob/main/README.md?plain=1#L253
It links to https://github.com/ucan-wg/spec/blob/main/README.md#54-rights-amplification instead of https://github.com/ucan-wg/spec#64-rights-amplification

nicobao:
Closes #115

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,7 +250,7 @@ Merging capability scopes MUST follow set semantics, where the result includes a
                  scope
 ```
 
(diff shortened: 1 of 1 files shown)


expede:
I really wish a "check if anchor is out of date" action existed 😭 There's ones that check for broken external links, but I can't find in-page ones. Seems like something that would be easy to build 🤔

Thanks for the fix!

PR #125, opened 2022-11-15
Testing a GH Action — please ignore

expede:


expede:
Well, TL;DR the `markdown-link-check` action still doesn't validate anchor tags, though the underlying library now does support that feature 🤷‍♀️

PR #126, opened 2022-11-16
fix: superuser ability

alanshaw:
As far as I can tell, `*` was removed in [this PR](https://github.com/ucan-wg/spec/pull/55/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L367) but _was_ top/super before it became `*/*`.

This PR updates the reference to be `*/*` and fixes the link to the relevant section in the spec.

Signed-off-by: Alan Shaw <alan.shaw@protocol.ai>

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ Abilities MUST NOT be case sensitive. For example, `http/post`, `http/POST`, `HT
 
 There MUST be at least one path segment as a namespace. For example, `http/put` and `db/put` MUST be treated as unique from each other.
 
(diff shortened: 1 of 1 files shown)


Gozala:
Wait is the `*/*` the thing now, I was not aware of this

expede:
@Gozala yeah, it was in the 0.9 PR in text and at least one diagram.

> does "" change to "/*" ?

Can you expand on this?

Gozala:
I have proposed to revert back to `*` instead in https://github.com/ucan-wg/spec/pull/127

Gozala:
we have reverted back to `*` instead.

PR #127, opened 2022-11-17
feat: reverts super user notation to *

Gozala:
Reverts super user notation from `*/*` back to `*`, with following justification

1. dag.house updated to 0.9 but with assumption that super user notation was `*`
   - Spec end up using `*/*` in few places but `*` in other, so we missed it.
   - Even if we add support for `*/*` we'd have to continue supporting `*`.
   - No one seems to have updated to `*/*` so maybe we can just revert to `*`.
2. notation `*/*` seems misleading because it leads to to believe `*/list` is also possible, in contract `*` doesn't invoke that misunderstanding

### DIFF
--- a/README.md
+++ b/README.md
@@ -529,12 +529,12 @@ If an attenuated resource or capability is desired, it MUST be explicitly listed
 
 ## 5.2 Top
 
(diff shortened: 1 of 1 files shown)


expede (approved):
LGTM thanks Irakli! Please also bump the spec version number to 0.9.2 :)

Issue #131, opened 2022-11-21
Syntax for an unattenuated delegation


Gozala:
Pulling this out of https://github.com/web3-storage/specs/pull/7/files#r1027388060. I would like to propose clear syntax for an unattenuated delegation between principals.

I am not sure if `"with": "own://did:key:zH3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV/*"` implies exactly that, if so it does not seem very intuitive, I would find `"with": "*"` far more intuitive and seems like it would align with `"can": "*"` too.

That said I think following would be most compact and most on point IMO:

```js
{
  iss: "did:key:zFrom",
  aud: "did:key:zTo",
  exp: null,
}
```

I would expect ☝️ to be equivalent of "also known as" in DID spec or elsewhere, implying that `did:key:zTo` could be used anywhere in place of `did:key:zFrom`.

Gozala:
>
>
>>    I mean yes, but I wish UCAN had a more explicit thing than relying on DID specifics
>
> Tell me more! Can you describe more about the kind of properties that you wish it had?

Does above description clarify it ?

In regards to DID specifics, I have hard time putting a finger on it. I suppose it is indirection. Explaining that ability to `did/rotate` implies principal has authority over all the capabilities of that `did` at hand is not easy and is somewhat distracting from the point itself which is `did:key:zTo` can be used in place of `did:key:zFrom` regardless of did document update takes place or not.

Gozala:
Ok I have a better explanation. In the context of https://github.com/web3-storage/specs/pull/7/files we would like some key to represent `did:dns:web3.storage` which is different from the key actual DID document will map to.

If we were to use `did/rotate` it would sort of imply that we are updating did key of the document to another key, but that is not what's happening. We are not actually rotating keys, or updating DID documents we just want to delegate complete authority to some other key.

expede:
# Clarity

> Does above description clarify it ?

_AHA!_ Oh wow okay this is what we've been talking past each other for many months about!

Yes, I see what you want now. The diagrams and wording are so similar. I think I was missing the pipelining idea.

# Pipelining

Aside from some lingering intuition about possible attacks, I think that your "pipeline" idea is _awesome._ If this were a "live"/dynamic ocap system, this would probably get modeled as a forwarder proxy. UCANs are "static", but I don't see why we can't express the same thing. You can always revoke the relevant UCAN.

[...]

Issue #130, opened 2022-11-20
Way to connect delegation chains


Gozala:
Pulling piece of discussion from https://github.com/web3-storage/w3protocol/issues/182#issuecomment-1321066496 after realizing that `did:dns` has nothing to do with actual problem.

Ok so let's consider following scenario:

1. User delegates some capability `did:key:zAlice` delegates to `did:key:zW3` service.
2. `did:key:zW3` private key is not in any computer it's on piece of paper in safe deposit box.
3. `did:key:zW3` has a delegate `did:key:zService` which acts on it's behalf
4. Service needs to redelegate capability received from `did:key:zAlice`, but it can not because delegation can not be signed by `did:key:zService`

In other words we have two delegation chains that if we were to connect we'd get a valid delegation chain. We could construct delegation like

```js
{
  iss: "did:key:zService",
  aud: "did:key:zZzzz",
  exp: null,
  att: [{ with: "did:key:zAlice", can: "*" }],
  prf: [
    // did:key:zAlice -> did:key:zW3
    {
      iss: "did:key:zAlice",
      aud: "did:key:zW3",
      exp: null,
      att: [{ with: "did:key:zAlice", can: "*" }],
    },
    // did:key:zW3 -> did:key:zService
    {
      iss: "did:key:zW3",
      aud: "did:key:zService",
      exp: null,
      att: [{ with: "*", can: "*" }],
    },
  ],
```

However above delegation would not be valid because:

1. We do not have `with: *` or a way to describe resources delegated to an issuer.
    - Perhaps `ucan:*` could be expanded to imply not only all proofs but also all siblings that delegate to the `iss` ?
    - Or we could make `with: *` be a thing ?
2. It is violates [principal alignment](https://github.com/ucan-wg/spec/#62-principal-alignment) because `aud` in second proof is `did:key:zW3` and not `did:key:zService`.
    - Perhaps spec should allow pushing misaligned proofs into a siblings with matching `iss`, because you are providing proof on their behalf. It would only work one level deep, but that seems ok.
    - Alternatively mismatched proofs could be moved into all the siblings, that way delegation several levels above could provide a proof on behalf of the issuer there.

Gozala:
I have wrote a draft of how we could address it (at least in specific context) https://github.com/web3-storage/specs/pull/7/files 

@expede would love your input, if you are able to take a look.

expede:
[...]

Issue #123, opened 2022-11-01
Multiple "can" for a "with"


oed:
Currently it seems like I would need to create multiple `with/can` statements for every action I want to allow for a resource. This means that I would add a lot of duplicate information if I want to delegate a lot of actions for a given resource.[ReCap](https://github.com/spruceid/EIPs/blob/19dae81ecc8354199bbe28f3b6248851f4cf49cf/EIPS/eip-5573.md#recap-details-object-schema) takes a slightly different approach to this were the resource is used as an object key and the value is an array of action strings. 

Maybe I'm missing some trick to represent multiple actions more efficiently?

matheus23:
Actions often imply each other. E.g. for CRUD actions, `can: DELETE` would imply `CREATE`, `READ`, and `UPDATE`, while `can: UPDATE` would imply `can: READ` but not `can: DELETE`.

In CRUD some sets can't be represented easily, e.g. a `CREATE, DELETE` set would need to be two distinct capabilities.
I don't think we've had a need for that in practice though. I guess the set of actions & their implications are just chosen appropriately to avoid having to provide multiple capabilities.

Gozala:
> I don't think we've had a need for that in practice though. I guess the set of actions & their implications are just chosen appropriately to avoid having to provide multiple capabilities.

I think that in by no means implied by the schema. If there is no need to have multiple actions on the same resource than I'd argue `att` schema should be more like:

```ts
"example://example.com/public/photos/": {
   can: "crud/delete",
   nb: {
      "matching": "/(?i)(\W|^)(baloney|darn|drat|fooey|gosh\sdarnit|heck)(\W|$)/"
   }
}
```

However I think ☝️ restrictions would be fairly impractical limitation & supporting multiple actions is inevitable. And in that case resources as keys would indeed be a lot more compact.

> And I would be delighted not to have to deal with `with` property which is [statement in JS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/with) and can't be a variable name.

However [ReCap](https://github.com/spruceid/EIPs/blob/19dae81ecc8354199bbe28f3b6248851f4cf49cf/EIPS/eip-5573.md#recap-details-object-schema) structure is not going to work either, as it does not allow for `nb`. However something along these lines would actually be really nice

```js
{
  "example://example.com/public/photos/": {
      "crud/delete": {}
  },
[...]

Issue #112, opened 2022-09-30
Under-specified format for revocation structure


cdata:
Regarding the section on revocation: https://github.com/ucan-wg/spec#66-revocation

A structure is prescribed to record a revocation:

```js
{
  "iss": did,
  "revoke": ucanCid,
  "challenge": sign(did.privateKey, `REVOKE:${ucanCid}`)
}
```

The types of the fields in this structure are not specified. It seems reasonable to assume that `did` is a DID string.

But, what is the form of `ucanCid`? A base64-encoded UCAN string? Or some other form?

And what is the form of the challenge?

matheus23:
Agreed.

Given `ucanCid` is re-used in a string splice & it's all JSON, let's assume `ucanCid` is base-something encoded that fits into ASCII, so either base32, base64, base58, base16? (Or should we rather standardize on what it is, so basically require base32 as the standard for now?)

The result of `sign` should probably be encoded as base64url, similar to how signatures in JWTs are encoded.

Also, `did.privateKey` seems to indicate that `did` is not a string, which may be confusing. We could rework this to say `didPrivateKey` or `privateKeyFor(did)`.

expede:
> But, what is the form of ucanCid? A base64-encoded CID string?

Yeah, fair.

## CIDs

I don't think that we can enforce a specific encoding of the CID without forcing a specific encoding of CIDs throughout the entire system. For purposes of caching revocations, the CID should be a normalized hash (likely to binary), rather than using the specifically text-encoded hash. This is the strategy in IPFS for caching, too. We could standardize on a specific hash function, since (e.g.) SHA2 and SHA3 will give a different hash for the same content.

## Signatures

We probably need to add the algorithm type to the signature, especially now that there's systems that use e.g. EIP-191.

## DIDs

The DID is a DID string, which is described in the relevant DID spec

expede:
Let's fix this on 0.10

expede:
Closed by #132

PR #132, opened 2022-11-23
v0.10

expede:
[Preview 📜](https://github.com/ucan-wg/spec/blob/0.10/README.md)

# Closes

* [x] Change `att` to `cap`?
* [x] Point all discussion of invocation at the [dedicated spec](https://github.com/ucan-wg/invocation/)
* [x] #112
* [x] #123
* [x] #131
* [x] #134
* [x] #135
* [x] #136
* [x] #141 [remove it]
* [x] #142
* [x] #143 
* [x] #145
* [x] Bump version to 0.10
* [x] Include https://github.com/ucan-wg/spec/issues/134
* [x] Move `ucv` field from header to body (to make implementation & e.g. CACAO interop easier)

# Removed/Deferred from Scope

* #130
  * By definition, this needs to live outside of the initial delegation
  * Several ways to do this (e.g. look in a DB, in a CAR file, some other format)
  * IMO this is super valuable, but is not in the minimal subset that everyone needs to support
  * As such it probably should have its own spec(s), removing from this PR
* #139
  * Discussed with several people
  * No clear benefit (e.g. entirely posisble to do a `did:mailto`)
  * Opening to URI is a mostly one-way door
    * i.e. It's much harder to reverse than add later if it's found to be critical


README.md
```diff
@@ -190,30 +192,25 @@ Further delegation of a capability with `nb` fields set MUST respect the `nb` fi
...
-    "with": "example://example.com/private/84MZ7aqwKn7sNiMGsSbaxsEa6EPnQLoKYbXByxNBrCEr",
-    "can": "wnfs/append"
-  },
-  {
-    "with": "example://example.com/public/photos/",
-    "can": "crud/delete",
-    "nb": {
-      "matching": "/(?i)(\W|^)(baloney|darn|drat|fooey|gosh\sdarnit|heck)(\W|$)/"
+    "example://example.com/public/photos/": {
+      "crud/delete": {},
```
Gozala, line 196:
```suggestion
      "crud/read": {},
```

I think having duplicate keys keys in the example is probably not a good idea, especially if we're saying we don't want to support it.

expede: Oh haha I think this is an actual typo from a copy/paste

README.md
```diff
@@ -190,30 +192,25 @@ Further delegation of a capability with `nb` fields set MUST respect the `nb` fi
 ``` json
 [
```
Gozala, line 193:
```suggestion
```

README.md
```diff
@@ -190,30 +192,25 @@ Further delegation of a capability with `nb` fields set MUST respect the `nb` fi
...
-      "templates": ["newsletter", "marketing"]
+    "mailto:username@example.com": {
+      "msg/send": {},
+      "msg/receive": {
+        "max_count": 5,
+        "templates": ["newsletter", "marketing"]
+      }
     }
   }
 ]
```
Gozala, line 216:
```suggestion
```

README.md
expede: No array?

expede: (Just loose objects?)

Gozala: Yeah I overlooked the details, my interpretation of discussion was https://github.com/ucan-wg/spec/pull/133

README.md
```diff
@@ -415,45 +403,111 @@ This field SHOULD NOT be used to sign arbitrary data, such as signature challeng
...
-This array also includes time ranges and the proofs that start latest and end soonest from the lower and upper time bounds of the range.
+#### 3.2.5.1 Anatomy
 
-### 3.2.7 Proof of Delegation
+The anatomy of a capability MUST be given as a mapping of resource URI to abilities to array of caveats or the wildcard `*`:
 
-Attenuations MUST be satisfied by matching the attenuated capability to a proof in the `prf` array ([§3.2.7.1](#3271-prf-field))
+```
+{ $RESOURCE: { $ABILITY: "*" } }
+{ $RESOURCE: { $ABILITY: [ ...$CAVEATS ] } }
```
expede, line 439: This may be controversial. I promise that there's a very good, very logical reason. See below.

README.md
```diff
@@ -415,45 +403,111 @@ This field SHOULD NOT be used to sign arbitrary data, such as signature challeng
...
+
+| Proof Caveats | Delegated Caveats | Is Valid? | Comment                              |
+|---------------|-------------------|-----------|--------------------------------------|
+| `"*"`         | `"*"`             | Yes       | Equal                                |
+| `[x]`         | `[x]`             | Yes       | Equal                                |
+| `[x]`         | `"*"`             | No        | Escalation to any                    |
+| `"*"`         | `[x]`             | Yes       | Introduces a new caveat              |
+| `[x, y]`      | `[x]`             | Yes       | Removes a capability                 |
+| `[x, y]`      | `[x, (y + z)]`    | Yes       | Attenuates existing caveat           |
+| `[x, y]`      | `[x, y, z]`       | No        | Escalation by adding new capability  |
```
expede, line 497: I ran into a problem. It was actually pretty counterintuitive to have `[]` mean "no caveats". Since we want to be able to express more than one _disjoint_ capabilty on the same resource/ability pair, this means that we have something less like a caveat and more like a capability or "opening" (i.e. matching "constructively").

Concretely, in UCAN 0.9 you can say this:

```json
[
  {
    "with": "https://example.com",
    "can": "crud/create",
    "nb": { "day": "monday", "status": "published" }
  },
  {
    "with": "https://example.com",
    "can": "crud/create",
    "nb": { "status": "draft" }
  }
]
```

i.e. you can create drafts at any time, but only publish on Mondays. In the new syntax:

```json
{
  "https://example.com": {
    "crud/create": [
      { "day": "monday", "status": "published" },
      {"status": "draft"}
    ]
  }
}
```

Which is itself straightforward. Now what does this mean?

```
```json
{
  "https://example.com": {
    "crud/create": []
  }
}
```

On the face of it, it means "no caveats". But this is strange! I can attenuate from `[a, b]` to `[a]`, which is a narrowing of permission scope. If I go from `[a]` to `[]`, it now _expands_ the scope! That's super counterintuitive! Here's the _old_ table:

| Proof Caveats | Delegated Caveats | Is Valid? | Comment |
|---------------|-------------------|-----------|--------|
| `[]`          | `[]`              | Yes       | Equal |
| `[a]`         | `[a]`             | Yes       | Equal |
| `[a]`         | `[]`              | No        | Escalation from some caveats to none |
| `[]`          | `[a]`             | Yes       | Introduces a new caveat |
| `[a, b]`      | `[a]`             | Yes       | Removes a capability     |
| `[a]`         | `[a, b]`          | Maybe     | Adds an attenuated capability |

The proposal in the current draft is to use `"*"` to mean "no caveats", and `[]` would mean "under no circumstances is this valid". If you look at the _new_ table (directly above this comment), it's much more consistent.

It's also more consistent with the `"*"` wildcard that we have in other places, such as `ucan:*`.

@Gozala and @dholms any thoughts?

expede: I also looked at `null` instead of `"*"`, but it was a bit confusing to read `"crud/create": null` as "unrestricted"

README.md
```diff
@@ -654,7 +716,7 @@ While some resources are centralized (e.g. access to a server), others are unbou
 
 Every resource type SHOULD have a canonical location where its revocations are kept. This list is non-exclusive, and revocation messages MAY be gossiped between peers in a network to share this information more quickly.
 
-It is RECOMMENDED that the canonical revocation store be kept as close to (or inside) the resource it is about as possible. For example, the WebNative File System maintains a Merkle tree of revoked CIDs at a well-known path. Another example is that a centralized server could have an endpoint that lists the revoked UCANs by CID.
+It is RECOMMENDED that the canonical revocation store be kept as close to (or inside) the resource it is about as possible. For example, the WebNative File System maintains a Merkle tree of revoked CIDs at a well-known path. Another example is that a centralized server could have an endpoint that lists the revoked UCANs by [canonical CID].
```
expede, line 719: This may be controversial, but I can't think of another way to make this work. 

To solve #112, we need a stable identifier. With something more general like ucan-ipld is that we have multiple posisble CIDs for the same data. This means that at minimum what gets signed over and CID need to match for the canonical CID. So that not every implementation has to support IPLD (or other formats e.g. RDF), we already require signing over a JWT (unless you don't care about consistency with other implementations, but then you're alread off-spec).

This means that the CID needs to be be pretty specific. CIDv1 + base32 + SHA256 seems like a pretty safe bet, but I'm absolutley open to other options.

@cdata do you have any opinions? In #112 you mention base64 — I'm happy to use that instead. Base32 means that you can pull up the CID in a URL via IPFS (for those using IPFS), but it's obvioulsy longer.

README.md
Gozala: How about we use `[{}]` to imply "no caveats" and treat `[]` as either:

1. Absence of delegation (which is I think what you're making a case for). In other words treat it same as `null` or as if key was not even there.
2. Alternatively it could mean inherit whatever caveats are in the proofs. I suspect this would be more error prone and probably not a good idea, but thought it was worth mentioning anyway.

I don't really like `"*"` because it introduces special string where maps are expected, `[{}]` on the other hand is basically what one would write if there were no caveats.

expede: Okay yeah `[{}]`is logical 👍

expede: Done!

README.md
Gozala: I'm not sure I understand the problem. Even though UCAN can be encoded in various formats issuer does sign over a payload in a specific encoding. Multihash of that payload is likely the best key for the revocation as it would not matter what the encoding of delegation is it's signature payload hash will remain the same.

Now that said unfortunately current UCAN format makes this bit trick and annoying unlike invocation format where task CID is explicitly captured inside an invocation. I have been wondering if we should do something similar with UCANs as well which would make revocation lookups simpler.

expede: The problem: since a UCAN can be respresented by multiple types of CID (e.g. CIDv0, CIDv1, etc), I have to at least parse every CID and extract just the hash, though this is potentially a lot of work on receipt of any UCAN. Restricting to a specific CID format, you can "just" match strings, but it feels a bit heavy-handed. I remember Juan once saying that CIDs aren't an interoperability mechanism between systems (to choose hashes etc), but rather a way to do upgrades with backwards compatability, and that in most cases applications should just stick with one CID format.

Gozala:
> Why is the `prf` field not a IPLD link? My assumption had been that it is, but looks like it's just a CID string.
> 
> Valid DagJSON would be:
> 
> ```json
> "prf": [
>   { "/": "bafkreihogico5an3e2xy3fykalfwxxry7itbhfcgq6f47sif6d7w6uk2ze" },
>   { "/": "bafkreiemaanh3kxqchhcdx3yckeb3xvmboztptlgtmnu5jp63bvymxtlva" }
> ]
> ```
> 
> Maybe there's a reason for this that I'm missing?

Historically UCANs were JWTs and all proofs were just inlined. With [UCAN-IPLD](https://github.com/ucan-wg/ucan-ipld/#24-proof) we introduced links in proofs although JWT version adopted strings.

I don't recall making a case for DAG-JSON links in JWT format, mainly because we use them in UCAN-IPLD in CBOR so it had no implications for us. I also expect that UCAN-IPLD in DAG-JSON format would use links.

While I'd prefer DAG-JSON like links in place of string encoded CIDs, it's probably up to JWT format users to make a informed decision on this.

bgins:
> While I'd prefer DAG-JSON like links in place of string encoded CIDs, it's probably up to JWT format users to make a informed decision on this.

After thinking this through following the WG meeting, I think this would be fairly confusing to JWT format users. They would have no context for the DAG-JSON format and would wonder why that syntax was chosen.

README.md
```diff
@@ -415,45 +463,114 @@ This field SHOULD NOT be used to sign arbitrary data, such as signature challeng
...
+| Proof Caveats | Comment                                                       |
+|---------------|---------------------------------------------------------------|
+| `[]`          | No capabilities                                               |
+| `[{}]`        | Full capabilities for this resource/ability pair (no caveats) |
+
+### 3.2.6 Proof of Delegation
+
+Attenuations MUST be satisfied by matching the attenuated capability to a proof in the [`prf` array][prf field].
+
+Proofs MUST be resolvable by the recipient. A proof MAY be left unresolvable if it is not used as support for the top-level UCAN's capability chain. The exact format MUST be defined in the relevant transport specification. Some examples of possible formats include: a JSON object payload delivered with the UCAN, a federated HTTP endpoint, a DHT, or shared database.
```
bgins, line 567:
```suggestion
Proofs MUST be resolvable by the recipient. A proof MAY be left unresolvable if it is not used as support for the top-level UCAN's capability chain. The exact format MUST be defined in the relevant transport specification. Some examples of possible formats include: a JSON object payload delivered with the UCAN, a federated HTTP endpoint, a DHT, or a shared database.
```

README.md
```diff
@@ -415,45 +463,114 @@ This field SHOULD NOT be used to sign arbitrary data, such as signature challeng
...
+}
 ```
 
-### 3.2.5 Attenuation
+### 3.2.5 Capabilities & Attenuation
 
-The attenuations (i.e. UCAN output, or "caveats") MUST be an array of heterogeneous capabilities (defined below). This array is REQUIRED but MAY be empty.
+Capabilities MUST be presented as a map. This map is REQUIRED but MAY be empty.
 
 This array MUST contain some or none of the following:
```
icidasset, line 488:
```suggestion
This map MUST contain some or none of the following:
```

README.md
```diff
@@ -515,54 +624,44 @@ The `ucan` scheme MUST accept the following ability: `ucan/*`. This ability rede
...
-    "with": "ucan:bafkreihogico5an3e2xy3fykalfwxxry7itbhfcgq6f47sif6d7w6uk2ze",
-    "can": "ucan/*"
-  }, 
-  { 
-    "with": "ucan:*", 
-    "can": "ucan/*" 
-  }
-]
+{ 
+  "ucan:bafkreihogico5an3e2xy3fykalfwxxry7itbhfcgq6f47sif6d7w6uk2ze": {"ucan:/*": []},
```
icidasset, line 628:
```suggestion
  "ucan:bafkreihogico5an3e2xy3fykalfwxxry7itbhfcgq6f47sif6d7w6uk2ze": {"ucan:/*": [{}]}, 
```

README.md
```diff
@@ -515,54 +624,44 @@ The `ucan` scheme MUST accept the following ability: `ucan/*`. This ability rede
...
-    "can": "ucan/*"
-  }, 
-  { 
-    "with": "ucan:*", 
-    "can": "ucan/*" 
-  }
-]
+{ 
+  "ucan:bafkreihogico5an3e2xy3fykalfwxxry7itbhfcgq6f47sif6d7w6uk2ze": {"ucan:/*": []}, 
+  "ucan:*": {"ucan/*": []}
```
icidasset, line 629:
```suggestion
  "ucan:*": {"ucan/*": [{}]}
```

README.md
```diff
@@ -471,38 +588,30 @@ Which in a JSON representation would resolve to the following table:
...
-scheme-selector = "*" / <scheme>
-```
-
-```
-own://did:key:zH3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV/*
-own://did:key:zH3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV/dns
-```
-
-The DID spec [permits an arbitrary DID method to specify the use of the URI path segment](https://www.w3.org/TR/did-core/#path), the `own` scheme separates out the reference of owned resources rather than a DID document itself. Other URI schemes (such as `dns` and `telnet`) define their own syntactic rules. `own` makes it clear that the contained DID is the "owner" of the resource in the path, fragments, and queries.
+`ucan:*` represents all of the UCANs in the current proofs array. If selecting a particular proof (i.e. not the wildcard), then the  MUST be used. In the case of selecting a particular proof, the validator MUST check that the delegated content address is listed in the proofs (`prf`) field.
```
bgins, line 614: > then the [?] MUST be used

It seems like there may be something missing here.

oed:
> I don't recall making a case for DAG-JSON links in JWT format, mainly because we use them in UCAN-IPLD in CBOR so it had no implications for us. I also expect that UCAN-IPLD in DAG-JSON format would use links.

So you have a manual conversion step I guess. Not using the native dag-json -> dag-cbor conversion logic?

bgins (commented):
Added a few copy edits.

README.md
expede: LOLOL thanks 👀

README.md
```diff
@@ -415,45 +463,114 @@ This field SHOULD NOT be used to sign arbitrary data, such as signature challeng
...
+Attenuations MUST be satisfied by matching the attenuated capability to a proof in the [`prf` array][prf field].
 
-#### 3.2.7.1 `prf` Field
+Proofs MUST be resolvable by the recipient. A proof MAY be left unresolvable if it is not used as support for the top-level UCAN's capability chain. The exact format MUST be defined in the relevant transport specification. Some examples of possible formats include: a JSON object payload delivered with the UCAN, a federated HTTP endpoint, a DHT, or a shared database.
 
-The `prf` field MUST contain the content address ([§6.5](#65-content-identifiers)) of UCAN proofs (the "inputs" of a UCAN). 
+#### 3.2.6.1 `prf` Field
 
-#### 3.2.7.2 Examples
+The `prf` field MUST contain the [content address][content identifiers] of UCAN proofs (the "inputs" of a UCAN).
```
matheus23, line 571: FWIW my initial reaction was that my vote would go out to the weird `{ "/": "<cid>" }` syntax for CIDs with dag-json.
The thinking was that we'd get e.g. `kubo pin` support and UCAN traversal via js-multiformats etc. for free, but that's not actually the case: The JWT encoding format with dotted and base64-ed payload + signature is in our way for that, so you'd need a canonicalization format like ucan-ipld in between anyways, so I'd say it's not worth the weirdness & extra bytes.

matheus23 (approved):
judges holding signs saying "10"

(UCAN version pun intended :stuck_out_tongue_winking_eye:)

README.md
```diff
@@ -163,74 +158,64 @@ Abilities MUST NOT be case sensitive. For example, `http/post`, `http/POST`, `HT
...
-  },
-  {
-    "with": "example://example.com/public/photos/",
-    "can": "crud/delete",
-    "nb": {
-      "matching": "/(?i)(\W|^)(baloney|darn|drat|fooey|gosh\sdarnit|heck)(\W|$)/"
-    }
+{
+  "example://example.com/public/photos/": {
+    "crud/read": [{}],
```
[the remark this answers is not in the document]
expede: Yeah I agree that this is confusing initially. I'm pretty sure that this syntax leaves open a couple ambiguities that have to get clarified in prose.

> is it more common to see AND-ed caveats than OR-ed caveats?

It is more common in contexts where you have a single capability at a time (which is arguably cleaner but less convenient especially in a passive setting)! The tradeoff is that it makes the syntax less expressive, because with an AND you can't express disjoint capabilities. I'm also unsure that your average programmer would know if it was an AND or an OR on first inspection — it's ambiguous in a way that the 0.9 syntax is not.

README.md
```diff
@@ -163,74 +158,64 @@ Abilities MUST NOT be case sensitive. For example, `http/post`, `http/POST`, `HT
...
-```
+## 2.5 Capability
 
-### 2.4.1 `nb` Non-Normative Fields
+A capability is the association of an "ability" to a "resource": `resource x ability x caveats`.
 
-Capabilities MAY define additional optional or required fields specific to their use case in the `nb` ([nota bene](https://en.wikipedia.org/wiki/Nota_bene)) field. This field is OPTIONAL in the general case, but MAY be REQUIRED by particular capability types that require this information to validate. The `nb` field MAY contain additional caveats or other important information related to specifying the capability, and MAY function as an "escape hatch" for when a use case is not fully captured by the `with` and `can` fields.
+The resource and ability fields are REQUIRED. Any non-normative extensions are OPTIONAL.
 
-Further delegation of a capability with `nb` fields set MUST respect the `nb` fields. On validation, if a `nb` field is present, it MUST be checked. If a validator is not able to interpret the `nb` field, it MUST reject the capability. As such, any `nb` caveats from a proof MUST further attenuate the delegated capability.
```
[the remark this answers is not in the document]
matheus23: Oh wait. Right. With the new syntax it's kind-of more obvious that you need to understand the full caveat before applying it.

Still, perhaps it'd be good to point out that let's say in this example
```json
{
  "https://example.com/blog": {
    "crud/update": [
      {"status": "published", "day-of-week": "Monday"} // only publish on Mondays
    ]
  }
}
```

If the parser understands caveats of the form `{"status": X}`, it MUST reject the caveat if there's other stuff in it. If it didn't and just ignored `day-of-week`, that'd be wrong.

I feel like that's the equivalent of the `nb` "MUST reject" sentence from 0.9.

README.md
```diff
@@ -247,120 +232,197 @@ Merging capability scopes MUST follow set semantics, where the result includes a
...
+The `iss` and `aud` fields describe the token's principals. These can be conceptualized as the sender and receiver of a postal letter. The token MUST be signed with the private key associated with the DID in the `iss` field. Implementations MUST include the [`did:key`] method, and MAY be augmented with [additional DID methods][DID].
 
-If an issuer's DID has more than one key (e.g. [`did:ion`](https://github.com/decentralized-identity/ion), [`did:3`](https://github.com/ceramicnetwork/CIP/blob/main/CIPs/CIP-79/CIP-79.md)), the key used to sign the UCAN MUST be made explicit, using the [DID fragment](https://www.w3.org/TR/did-core/#fragment) (the hash index).
+The `iss` and `aud` fields MUST contain a single principal each.
 
-The underlying key types RSA, ECDSA, and EdDSA MUST be supported. Use of ECDSA is supported but [RECOMMENDED that another key type be preferred](https://en.wikipedia.org/wiki/Elliptic_Curve_Digital_Signature_Algorithm#Security).
+If an issuer's DID has more than one key (e.g. [`did:ion`], [`did:3`]), the key used to sign the UCAN MUST be made explicit, using the [DID fragment] (the hash index).
```
dholms, line 423: to verify: `aud` _does not_ have to specify a key using a hash fragment right?

expede: Correct! I can expand the text to clarify :)

README.md
```diff
@@ -247,120 +232,197 @@ Merging capability scopes MUST follow set semantics, where the result includes a
...
+
+The underlying key types RSA, ECDSA, and EdDSA MUST be supported. Use of ECDSA is supported but [RECOMMENDED that another key type be preferred][ECDSA security].
```
dholms, line 425: Does RSA _have_ to be supported? We don't support RSA keys generally for instance 🤔 

(we actually don't support EdDSA yet either but will soon 😅)

README.md
```diff
@@ -413,45 +475,114 @@ This field SHOULD NOT be used to sign arbitrary data, such as signature challeng
...
+
+| Proof Caveats | Delegated Caveats | Is Valid? | Comment                              |
+|---------------|-------------------|-----------|--------------------------------------|
+| `[{}]`        | `[{}]`            | Yes       | Equal                                |
+| `[x]`         | `[x]`             | Yes       | Equal                                |
+| `[x]`         | `[{}]`            | No        | Escalation to any                    |
+| `[{}]`        | `[x]`             | Yes       | Introduces a new caveat              |
+| `[x, y]`      | `[x]`             | Yes       | Removes a capability                 |
+| `[x, y]`      | `[x, (y + z)]`    | Yes       | Attenuates existing caveat           |
+| `[x, y]`      | `[x, y, z]`       | No        | Escalation by adding new capability  |
```
dholms, line 568: I think a line in here about 
`[x] -> [y]` which would not be allow unless `y` is an attenuation of `x`

I also might slightly change the wording around
`[{}] -> [x]` because you are not "introducing a new caveat", you are "attenuating `[{}]` to `[x]`"

README.md
```diff
@@ -585,26 +698,37 @@ A UCAN is valid inclusive from the `nbf` time and until the `exp` field. If the
 
 In delegation, the `aud` field of every proof MUST match the `iss` field of the outer UCAN (the one being delegated to). This alignment MUST form a chain back to the originating principal for each resource.
```
dholms, line 699: Related to my prev comment: does this hold if the `iss` is a did that points to a key with the hash fragment?

README.md
```diff
@@ -663,8 +796,8 @@ A revocation message MUST conform to the following format:
 ``` js
 {
   "iss": did,
-  "revoke": ucanCid,
-  "challenge": sign(did.privateKey, `REVOKE:${ucanCid}`)
+  "revoke": canonicalUcanCid,
+  "challenge": sign(did.privateKey, `REVOKE:${canonicalUcanCid}`)
```
dholms, line 800: how is this json-encoded? base64? multibase?

expede: Good point! base64 👍 I'll add that to the text

README.md
expede: Yup! The hash fragment is just there to help figure out the specific key that signed, which is different from which agent is being delegated to. Will add clarifying text 👍

README.md
```diff
@@ -696,7 +696,9 @@ A UCAN is valid inclusive from the `nbf` time and until the `exp` field. If the
 
 ## 6.2 Principal Alignment
 
-In delegation, the `aud` field of every proof MUST match the `iss` field of the outer UCAN (the one being delegated to). This alignment MUST form a chain back to the originating principal for each resource.
+In delegation, the `aud` field of every proof MUST match the `iss` field of the outer UCAN (the one being delegated to). This alignment MUST form a chain back to the originating principal for each resource. 
+
+This calculation MUST NOT take into account [DID fragment]s. If present, fragments are only intended to clarify which of a DID's keys was used to sign a particular UCAN, not to limit which specific key is delegated between. Use of `did:key` if delegation to a specific key is desired.
```
dholms, line 702: Think a typo here? (unless im reading it wrong)

> Use of `did:key` if delegation to a specific key is desired.

Remove "of"

dholms (approved):
This all looks great to me! 🙌 

Thanks for the quick back & forth on a couple things!

bgins (approved):
Looks great! This spec has come out really nicely. ✨ Thanks for all of the work pulling our voices together!

icidasset (approved):

README.md
```diff
@@ -515,54 +639,44 @@ The `ucan` scheme MUST accept the following ability: `ucan/*`. This ability rede
...
-    "with": "ucan:bafkreihogico5an3e2xy3fykalfwxxry7itbhfcgq6f47sif6d7w6uk2ze",
-    "can": "ucan/*"
-  }, 
-  { 
-    "with": "ucan:*", 
-    "can": "ucan/*" 
-  }
-]
+{ 
+  "ucan:bafkreihogico5an3e2xy3fykalfwxxry7itbhfcgq6f47sif6d7w6uk2ze": {"ucan:/*": [{}]},
```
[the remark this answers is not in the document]
expede: Thanks @hugomrdias!

README.md
Gozala: What I meant to say is if UCAN had a following structure:

```haskell
type struct UCAN {
  payload: Link<UCANPayload>
  sig: VarSig
}
```

It would not matter what format you got the actual UCAN in because issuer would have signed over whatever the link in whatever encoding they chose. It does mean that you're facing that you may now get arbitrary hashes and formats, but if you don't support those algorithms well you won't be able to verify those UCANs anyway so revocations are probably not a concern, if you do support them you're back to comparing stings. And because we embraced CIDs base encoding is no longer relevant. Recepient can transcode key into base encoding that matches their store encoding of keys.

README.md
```diff
@@ -515,54 +639,44 @@ The `ucan` scheme MUST accept the following ability: `ucan/*`. This ability rede
...
-    "can": "ucan/*"
-  }, 
-  { 
-    "with": "ucan:*", 
-    "can": "ucan/*" 
-  }
-]
+{ 
+  "ucan:bafkreihogico5an3e2xy3fykalfwxxry7itbhfcgq6f47sif6d7w6uk2ze": {"ucan/*": [{}]}, 
+  "ucan:*": {"ucan/*": [{}]}
```
Gozala, line 644: Do we need `ucan/` prefix on the ability here ? We used to have special `*` ability to denote "all" abilities which was far more intuitive than `ucan/*`. If it really need to be name-spaced I would  suggest `*/*` as it feels far more intuitive that `*` and `*/*` would include `crud/read` than `ucan/*`.

Gozala: We also use `{ with: "ucan:*", can: "*" }` when linking devices which I think would be considered invalid ? If so it is probably worth calling out what it would imply.

README.md
```diff
@@ -515,54 +639,44 @@ The `ucan` scheme MUST accept the following ability: `ucan/*`. This ability rede
...
+    crud/mutate --> crud/*
+
+    subgraph mutationGraph [ ]
+        crud/create --> crud/mutate
+        crud/update --> crud/mutate
+        crud/destroy --> crud/mutate
+    end
+  end
+
+  ... --> *
```
Gozala, line 676: I would really like if the hierarchy above was actually part of the spec, meaning that it would be really useful if it was specified that `msg/*` includes `msg/send` and `*` included `msg/*`  and `crud/*` etc.. Which as I understood from our side channel conversation is not implied

Gozala (approved):
Looks good to me. I do wish we could resolve some of the ability taxonomy, but perhaps a next revision where it should occur.

burdiyan:
Love the new structure for attenuation! Feels much better than previous list of `with` and `can`. I guess this addresses my comment in the [Discussion section](https://github.com/orgs/ucan-wg/discussions/122#discussioncomment-4704375).

README.md
```diff
@@ -471,38 +603,30 @@ Which in a JSON representation would resolve to the following table:
...
+|-------------------------|-----------------------------------------------|
+| `ucan:<cid>`            | A specific UCAN by [CID][content identifiers] |
+| `ucan:*`                | All possible provable UCANs                   |
+| `ucan:./*`              | All in this UCAN's proofs                     |
+| `ucan://<did>/*`        | All of any scheme "owned" by a DID            |
+| `ucan://<did>/<scheme>` | All of scheme "owned" by a DID                |
 
-The DID spec [permits an arbitrary DID method to specify the use of the URI path segment](https://www.w3.org/TR/did-core/#path), the `own` scheme separates out the reference of owned resources rather than a DID document itself. Other URI schemes (such as `dns` and `telnet`) define their own syntactic rules. `own` makes it clear that the contained DID is the "owner" of the resource in the path, fragments, and queries.
+`ucan:*` represents all of the UCANs in the current proofs array. If selecting a particular proof (i.e. not the wildcard), then its CID MUST be used. In the case of selecting a particular proof, the validator MUST check that the delegated content address is listed in the proofs (`prf`) field.
```
icidasset, line 629: @expede Bit confused by this section. This paragraph says:

> `ucan:*` represents all of the UCANs in the current proofs array.

But the table above says:

> `ucan:*` All possible provable UCANs
> `ucan:./*`All in this UCAN's proofs

Does the syntax in the paragraph need to be adjusted to `ucan:./*`?
Or am I confused about what `ucan:*` actually means?
I assume it means more than just the current proofs array?

README.md
expede: At the level of the UCAN spec, we're leaving it by convention for now. Not all namespaces have a `.../*`, but I do think that we want to start writing related specs for essentially a stdlib of common abilities and resources 💯

README.md
expede: Ah indeed, we should make this consistent — apologies. It's also something that I don't _love_ about this syntax: it's not very salient despite the meaning being quite different

PR #133, opened 2022-11-25
update example as per discussion in #123

Gozala:
Signed-off-by: Irakli Gozalishvili <contact@gozala.io>

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,30 +190,31 @@ Further delegation of a capability with `nb` fields set MUST respect the `nb` fi
 #### 2.4.2 Examples
 
 ``` json
-[
-  {
-    "example://example.com/public/photos/": {
-      "crud/read": {},
-      "crud/delete": {
-        "matching": "/(?i)(\\W|^)(baloney|darn|drat|fooey|gosh\\sdarnit|heck)(\\W|$)/" 
+{
+  "example://example.com/public/photos/": {
+    "crud/read": [],
+    "crud/delete": [
+      {
+        "matching": "/(?i)(\\W|^)(baloney|darn|drat|fooey|gosh\\sdarnit|heck)(\\W|$)/"
       }
-    }
+    ]
   },
-  {
-    "example://example.com/private/84MZ7aqwKn7sNiMGsSbaxsEa6EPnQLoKYbXByxNBrCEr": { 
-      "wnfs/append": {} 
-    }
+  "example://example.com/private/84MZ7aqwKn7sNiMGsSbaxsEa6EPnQLoKYbXByxNBrCEr": {
+    "wnfs/append": []
   },
-  {
-    "mailto:username@example.com": {
-      "msg/send": {},
-      "msg/receive": {
+  "mailto:username@example.com": {
+    "msg/send": [],
+    "msg/receive": [
+      {
         "max_count": 5,
-        "templates": ["newsletter", "marketing"]
+        "templates": [
+          "newsletter",
+          "marketing"
+        ]
       }
-    }
+    ]
   }
-]
+}
 ```
 
 ## 2.5 Capability Scope


expede:
@Gozala ah, okay, you want to enforce key uniqueness. What is the expected behaviour if they include duplicate keys? Is it indeterminate, or do we want to enforce an extra validation step?

Gozala:
> @Gozala ah, okay, you want to enforce key uniqueness. What is the expected behaviour if they include duplicate keys? Is it indeterminate between parsers, or do we want to enforce an extra validation step?

From what I can tell keys must be unique per spec

image

In JS `JSON.parse` will drop duplicated keys and just use the last one. I think these are strong enough reasons to expect that duplicate keys will not be really used & spec could note that this case is not supported.

expede:
@Gozala they "SHOULD" be unique, not MUST be unique. From the same spec:

```
    When the names within an object are not
    unique, the behavior of software that receives such an object is
    unpredictable. Many implementations report the last name/value pair
    only. Other implementations report an error or fail to parse the object,
    and some implementations report all of the name/value pairs,
    including duplicates.

    JSON parsing libraries have been observed to differ as to whether or
    not they make the ordering of object members visible to calling
    software. Implementations whose behavior does not depend on member
    ordering will be interoperable in the sense that they will not be
    affected by these differences.
```

Gozala:
I think copying that exact segment would be reasonable here as well

> When the names within an object are not unique, the behavior of software that receives such an object is unpredictable.

Gozala:
Also I don't believe spec states anywhere that `att`, `prf`, `can`, `with` or any other fields MAY not be duplicated it's kind of implied, so I'm not sure why this specific one would have to be explicitly called out.

expede:
I'm not dead set against this, but we did have someone do this in practice previously (which is why I know that section of the spec). It is a Byzantine condition (buggy not malicious) that is "yet another small detail that implementers can mess up".

I agree that it reads more nicely without the extra wrapping, but also is it worth opening the door for more broken certificates in the wild in exchange for a handful of characters? I genuinely don't know!

expede:
> I'm not sure why this specific one would have to be explicitly called out.

Because this one is much easier to mess up. You're often taking input of "which capsbilities (plural)". You never take a list of expirations.

expede:
But again — if we're willing to say "some UCANs will be read as invalid even if they're not actually, and that may not be the fault of the latest user in the chain", then we can add this constraint to the spec 👍 We just need to do it intentionally

expede (approved):
@Gozala https://github.com/ucan-wg/spec/pull/133#issuecomment-1327939197 and we can tick the box of "we thought about it", I'm happy to sign off on this change :)

Gozala:
> I agree that it reads more nicely without the extra wrapping, but also is it worth opening the door for more broken certificates in the wild in exchange for a handful of characters? I genuinely don't know!

Point I'm trying to make that this is not making it necessarily worse, because you just as well could encounter something like this today:

```ts
[
  {
     with: "did:key:zAlice",
     can: "inbox/read",
     can: "inbox/write"
  }
]
```

I do not believe behavior of the above is specified & JS implementation at least will simply discard `can: "index/read"`.

In other words supporting duplicate keys would be significantly complicated in some languages, so perhaps spec should state that keys MUST be unique. Even than I doubt JS implementation would handroll JSON parser to to verify that no duplicate keys are present.

expede:
>  because you just as well could encounter something like this today:

Right, you could run into that! It's much less likely because of how this stuff gets tends to get passed around

expede:
> In other words supporting duplicate keys would be significantly complicated in some languages, so perhaps spec should state that keys MUST be unique.

Yeah agreed: if we do this, then we have to specify that the keys MUST be unique

expede:
(I've had this conversation with a few people the past couple weeks, so I'm comfortable with what would have to happen here if this is the direction we go. Merge if and when ready — it's pre-approved above)

Gozala:
> Right, you could run into that! It's much less likely because of how this stuff gets tends to get passed around

I think what I'm hearing that implementation are more likely to add a resource to the `att` without checking if key for the resource exists. Which in some languages will produce duplicate keys while in others it would simply override pre-existing resource field, in both cases neither would be intended behavior.

I personally find this to be an ok tradeoff. Spec can warn about this, compliance tests could enforce this and I expect implementation will be able to avoid this. This does not fully eliminate possibility that such UCANs may be produced, yet I estimate probability to be low enough that improved readability is worth it.

Also worth a note that IPLD does not support duplicate keys either.

expede:
> Also worth a note that IPLD does not support duplicate keys either.

Yup! dag-json is extremely constrained compared to JSON!

expede:
> yet I estimate probability to be low enough that improved readability is worth it

I personally don't really find them to read too differently, but also they're so close that it's splitting hairs. I'm comfortable with the `MUST` clause (and we still have lots of time to decide against it if some horrible edge case comes up before we bump to 0.10)

PR #149, opened 2023-02-10
typo

nichoth:
Wasn't sure if this was an error. The text as it was written did not make sense to me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -648,7 +648,7 @@ The resolution of these addresses is left to the implementation and end-user, an
 
 Any issuer of a UCAN MAY later revoke that UCAN or the capabilities that have been derived from it further downstream in a proof chain.
 
-This mechanism is eventually consistent and SHOULD be considered the last line of defense against abuse. Proactive expiry via time bounds or other constraints SHOULD be preferred, as they do not require learning more information than what would be available on an offline computer.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -648,7 +648,7 @@ The resolution of these addresses is left to the implementation and end-user, an
 
 Any issuer of a UCAN MAY later revoke that UCAN or the capabilities that have been derived from it further downstream in a proof chain.
 
-This mechanism is eventually consistent and SHOULD be considered the last line of defense against abuse. Proactive expiry via time bounds or other constraints SHOULD be preferred, as they do not require learning more information than what would be available on an offline computer.
+This mechanism is eventually consistent and SHOULD be considered the last line of defense against abuse. Proactive expiry via time bounds or other constraints SHOULD be preferred, as they do not require learning more information than what would not be available on an offline computer.
```
matheus23, line 651: To me the original wording makes sense:
The information available on an offline computer that is referred to here is just system clock time: It can be used to check whether a UCAN is expired, even when the computer checking it is offline.
Additionally the paragraph is a little more lax about only using time bounds as constraints, as perhaps there are other constraints that would work on offline computers.

nichoth: Thanks I understand now

matheus23: This was good feedback though. I agree that the sentence isn't easy to understand and perhaps there's a way of writing it that wouldn't have been confusing the first time.

PR #161, opened 2023-07-12
Clarify that timestamp are integer seconds

### LINKED ISSUE #136: Clarify timestamp percision in the spec
Recently I noticed that spec refers to timestamps without mentioning precision.

https://github.com/ucan-wg/spec/blob/692e8aab59b763a783fe1484131c3f40d997b69a/README.md?plain=1#L350-L351

Implementations on the other hand seem to use seconds based precision and encode as such in JWT as well.

I think spec should clarify what precision supposed to be used.

expede:
- Closes #136


expede:
This is such a minor change, and was mostly already addressed in 0.10. I'm just going to merge this and people can complain later 😛

PR #165, opened 2023-08-02
Fix broken anchor in multicodec `0x55` link

### LINKED ISSUE #164: Update raw data multicodec link
The link in the spec currently points to the `protobuf` entry in the multicodec table: https://github.com/ucan-wg/spec/blob/b172154d76d893a7fdb9376d7c28800ee79b0965/README.md?plain=1#L1080

Likely the entries in the table shifted around. We should update the link or maybe avoid referencing by line.

expede:
Closes #164 

Using a permalink per @bumblefudge's suggestion. I don't think it's a huge problem if the table updates, especially since removing entries from Multocodec is so rare.


bgins (approved):
Permalink seems good to me. 👌

bumblefudge:
> especially since removing entries from Multocodec is so rare.

hold my beer

PR #173, opened 2023-10-09
Top-Level Spec

expede:
[Preview 📚](https://github.com/ucan-wg/spec/blob/high-level/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,7 +332,7 @@ For example, a capability may used to represent the ability to send email from a
 >
 > —[Mark Miller], [Robust Composition]
 
-The set of capabilities delegated by a UCAN is called its "authority." To frame it another way, it's the set of effects that a principal can cause, and functions as a declarative description of delegated abilities.
+The set of capabilities delegated by a UCAN is called its "authority." To frame it another way, it's the set of effects that a principal can cause, and acts as a declarative description of delegated abilities.
 
 Merging capability authorities MUST follow set semantics, where the result includes all capabilities from the input authorities. Since broader capabilities automatically include narrower ones, this process is always additive. Capability authorities can be combined in any order, with the result always being at least as broad as each of the original authorities.
 
--- a/README.md
+++ b/README.md
@@ -22,17 +22,17 @@
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [BCP 14] when, and only when, they appear in all capitals, as shown here.
 
-# 0. Abstract
+# Abstract
 
 User-Controlled Authorization Network (UCAN) is a [trustless], secure, [local-first], user-originated, distributed authorization scheme. This document provides a high level overview of the components of the system, concepts, and motivation. Exact formats are given in [sub-specifications].
 
-# 1. Introduction
+# Introduction
 
 User-Controlled Authorization Network (UCAN) is a [trustless], secure, [local-first], user-originated, distributed authorization scheme. It provides public-key verifiable, delegable, expressive, openly extensible [capabilities]. UCANs achieve public verifiability with late-bound certificate chains and principals represented by [decentralized identifiers (DIDs)][DID].
 
 UCAN improves the familiarity and adoptability of schemes like [SPKI/SDSI][SPKI] for web and native application contexts. UCAN allows for the creation, delegation, and invocation of authority by any agent with a DID, including traditional systems and peer-to-peer architectures beyond traditional cloud computing.
 
-## 1.1 Motivation
+## Motivation
 
 > If we practice our principles, we could have both security and functionality. Treating security as a separate concern has not succeeded in bridging the gap between principle and practice, because it operates without knowledge of what constitutes least authority.
 >
@@ -50,27 +50,27 @@ Many high-value applications run in hostile environments. In recognition of this
 
 Two related models that work exceptionally well in the above context are Simple Public Key Infrastructure ([SPKI][spki rfc]) and object capabilities ([OCAP]). Since offline operation and self-verifiability are two requirements, UCAN adopts a [certificate capability model] related to [SPKI].
 
-## 1.2 Intuition for Auth System Differences 
+## Intuition for Auth System Differences 
 
 The following analogies illustrate several significant trade-offs between these systems but are only accurate enough to build intuition. A good resource for a more thorough presentation of these trade-offs is [Capability Myths Demolished]. In this framework, UCAN approximates SPKI with some dynamic features.
 
-### 1.2.1 Access Control Lists
+### Access Control Lists
 
 By analogy, ACLs are like a bouncer at an exclusive event. This bouncer has a list attendees allowed in and which of those are VIPs that get extra access. People trying to get in show their government-issued ID and are accepted or rejected. In addition, they may get a lanyard to identify that they have previously been allowed in. If someone is disruptive, they can simply be crossed off the list and denied further entry.
 
 If there are many such events at many venues, the organizers need to coordinate ahead of time, denials need to be synchronized, and attendees need to show their ID cards to many bouncers. The likelihood of the bouncer letting in the wrong person due to synchronization lag or confusion by someone sharing a name is nonzero.
 
-### 1.2.2 Certificate Capabilities
+### Certificate Capabilities
 
 UCANs work more like [movie tickets][caps as keys] or a festival pass. No one needs to check your ID; who you are is irrelevant. For example, if you have a ticket issued by the theater to see Citizen Kane, you are admitted to Theater 3. If you cannot attend an event, you can hand this ticket to a friend who wants to see the film instead, and there is no coordination required with the theater ahead of time. However, if the theater needs to cancel tickets for some reason, they need a way of uniquely identifying them and sharing this information between them.
 
-### 1.2.3 Object Capabilities
+### Object Capabilities
 
 Object capability ("ocap") systems use a combination of references, encapsulated state, and proxy forwarding. As the name implies, this is fairly close to object-oriented or actor-based systems. Object capabilities are [robust][Robust Composition], flexible, and expressive.
 
 To achieve these properties, object capabilities have two requirements: [fail-safe], and locality preservation. The emphasis on consistency rules out partition tolerance[^pcec].
 
-## 1.3 Security Considerations
+## Security Considerations
 
 Each UCAN includes an assertions of what it is allowed to do. "Proofs" are positive evidence (elsewhere called "witnesses") of the possession of rights. They are cryptographically verifiable chains showing that the UCAN issuer either claims to directly own a resource, or that it was delegated to them by some claimed owner. In the most common case, the root owner's ID is the only globally unique identity for the resource.
 
@@ -80,7 +80,7 @@ Note that a structurally and cryptographically valid UCAN chain can be semantica
 
 While certificate chains go a long way toward improving security, they do not provide [confinement] on their own. The principle of least authority SHOULD be used when delegating a UCAN: minimizing the amount of time that a UCAN is valid for and reducing authority to the bare minimum required for the delegate to complete their task. This delegate should be trusted as little as is practical since they can further sub-delegate their authority to others without alerting their delegator. UCANs do not offer confinement (as that would require all processes to be online), so it is impossible to guarantee knowledge of all of the sub-delegations that exist. The ability to revoke some or all downstream UCANs exists as a last resort.
 
-## 1.4 Inversion of Control
+## Inversion of Control
 
 [Inversion of control] is achieved due to two properties: self-certifying delegation and reference passing. There is no Authorization Server (AS) that sits between requestors and resources. In traditional terms, the owner of a UCAN resource is the resource server (RS) directly.
 
@@ -120,7 +120,7 @@ This inverts the usual relationship between resources and users: the resource gr
 
 This additionally allows UCAN to model auth for [eventually consistent and replicated state][overcoming SSI].
 
-# 2 Lifecycle
+# Lifecycle
 
 The UCAN lifecycle has three parts:
 
@@ -148,7 +148,7 @@ flowchart TD
     click rev href "https://github.com/ucan-wg/revocation" "UCAN Revocation Spec"
 ```
 
-## 2.1 Time
+## Time
 
 It is often useful to talk about a UCAN in the context of some action. For example, a UCAN delegation may be valid when it was created, but expired when invoked.
 
@@ -163,30 +163,30 @@ sequenceDiagram
     Alice ->> Alice: Execute
 ```
 
-### 2.1.1 Validity Interval
+### Validity Interval
 
 The period of time that a capability is valid from and until. This is the range from the latest "not before" to the earliest expiry in the UCAN delegation chain.
 
-### 2.1.2 Delegation-Time
+### Delegation-Time
 
 The moment at which a delegation is asserted. This MAY be captured by an `iat` field, but is generally superfluous to capture in the token.
 
-### 2.1.3 Invocation-Time
+### Invocation-Time
 
 The moment a UCAN Invocation is created. It must be within the Validity Interval.
 
-### 2.1.4 Validation-Time
+### Validation-Time
 
 Validation MAY occur at multiple points during a UCAN's lifecycle. The main two are:
 
 - On receipt of a delegation
 - When executing an invocation
 
-### 2.1.5 Execution-Time
+### Execution-Time
 
 To avoid the overloaded word "runtime", UCAN adopts the term "execution-time" to express the moment that the executor attempts to use the authority captured in an invocation and associated delegation chain. Validation MUST occur at this time.
 
-## 2.2 Example
+## Example
 
 Here is a concrete example of all stages of the UCAN lifecycle for database write access.
 
@@ -218,9 +218,9 @@ sequenceDiagram
     DBAgent -X Bob: NAK(➏) [rejected]
 ```
 
-# 3. Terminology
+# Terminology
 
-## 3.1 Roles
+## Roles
 
 There are several roles that an agent MAY assume:
 
@@ -265,7 +265,7 @@ flowchart TD
     end
 ```
 
-## 3.2 Subject
+## Subject
 
 > At the very least every object should have a URL
 >
@@ -277,7 +277,7 @@ flowchart TD
 
 A Subject MUST be referenced by [DID]. This behaves much like a [GUID], with the addition of public key verifiability. This unforgeability prevents malicious namespace collisions which can lead to [confused deputies][confused deputy problem].
 
-## 3.3 Resource
+## Resource
 
 A resource is some data or process that can be uniquely identified by a [URI]. It can be anything from a row in a database, a user account, storage quota, email address, etc. Resource MAY be as coarse or fine grained as desired. Finer-grained is RECOMMENDED where possible, as it is easier to model the principle of least authority ([PoLA]).
 
@@ -287,7 +287,7 @@ Having a unique agent represent a resource (and act as its manager) is RECOMMEND
 
 Unless explicitly stated, the Resource of a UCAN MUST be the Subject.
 
-## 3.4 Command
+## Command
 
 Commands are concrete messages ("verbs") that MUST be unambiguously interpretable by the Subject of a UCAN. Commands are REQUIRED in invocations. Some examples include `msg/send`, `crud/read`, and `ucan/revoke`.
 
@@ -299,7 +299,7 @@ Commands MUST NOT be case-sensitive. There MUST be at least one path segment as 
 
 The slash convention (`my/command`, `my/other/command`) is only that: a namespacing _convention_. `dostuff`, `my/special/command` and `bbq-case-command` MUST all be accepted. The slash convention is RECOMMENDED to avoid namespace collision.
 
-### 3.4.1 Ability
+### Ability
 
 Abilities abstract over [Command]s to allow for extension of UCAN delegations.
 
@@ -307,11 +307,11 @@ Abilities MAY be organized in a hierarchy. A typical example is a superuser capa
 
 Abilities MUST NOT be case-sensitive. There MUST be at least one path segment as a namespace. For example, `http/put` and `db/put` MUST be treated as unique from each other.
 
-## 3.5 Conditions
+## Conditions
 
 Capabilities MAY define additional optional or required fields specific to their use case in the Condition fields. This field is OPTIONAL in the general case, but MAY be REQUIRED by particular capability types that require this information to validate. Conditions MAY function as an "escape hatch" for when a use case is not fully captured by the resource and ability fields.
 
-## 3.6 Capability
+## Capability
 
 A capability is the association of an ability to a subject: `subject x command x arguments x conditions`.
 
@@ -326,7 +326,7 @@ For example, a capability may used to represent the ability to send email from a
 | Arguments  | `{sender: "mailto:alice@example.com"}`                                       |
 | Conditions | `[{"day": "friday"}, {"field": "to", "includes": "mailto:bob@example.com"}]` |
 
-## 3.7 Authority
+## Authority
 
 > Whether to enable cooperation or to limit vulnerability, we care about _authority_ rather than _permissions._ Permissions determine what actions an individual program may perform on objects it can directly access. Authority describes the effects that a program may cause on objects it can access, either directly by permission, or indirectly by permitted interactions with other programs.
 >
@@ -363,11 +363,11 @@ Merging capability authorities MUST follow set semantics, where the result inclu
 
 The capability authority is the total rights of the authorization space down to the relevant volume of authorizations. Individual capabilities MAY overlap; the authority is the union. Every unique delegated capability MUST have equal or narrower capabilities from their delegator. Inside this content space, you can draw a boundary around some resource(s) (their type, identifiers, and paths or children) and their capabilities.
 
-## 3.8 Attenuation
+## Attenuation
 
 Attenuation is the process of constraining the capabilities in a delegation chain. Each direct delegation MUST either directly restate or attenuate (diminish) its capabilities.
 
-# 4. Token Resolution
+# Token Resolution
 
 Token resolution is transport specific. The exact format is left to the relevant UCAN transport specification. At minimum, such a specification MUST define at least the following:
 
@@ -377,33 +377,33 @@ Token resolution is transport specific. The exact format is left to the relevant
 
 Note that if an instance cannot dereference a CID at runtime, the UCAN MUST fail validation. This is consistent with the [constructive semantics] of UCAN.
 
-## 4.1 Content Identifiers
+## Content Identifiers
 
 A UCAN token SHOULD be referenced as a [base32][multibase] [CIDv1]. [SHA2-256] is the RECOMMENDED hash algorithm. The [DAG-CBOR] codec MUST be supported, and [DAG-JSON] support is RECOMMENDED. 
 
 The resolution of these addresses is left to the implementation and end-user, and MAY (non-exclusively) include the following: local store, a distributed hash table (DHT), gossip network, or RESTful service.
 
-# 5. Implementation Recommendations
+# Implementation Recommendations
 
-## 5.1 Delegation Store
+## Delegation Store
 
 A validator MAY keep a local store of UCANs that it has received. UCANs are immutable but also time-bound so that this store MAY evict expired or revoked UCANs.
 
 This store SHOULD be indexed by CID (content addressing). Multiple indices built on top of this store MAY be used to improve capability search or selection performance.
 
-## 5.2 Memoized Validation
+## Memoized Validation
 
 Aside from revocation, capability validation is idempotent. Marking a CID (or capability index inside that CID) as valid acts as memoization, obviating the need to check the entire structure on every validation. This extends to distinct UCANs that share a proof: if the proof was previously reviewed and is not revoked, it is RECOMMENDED to consider it valid immediately.
 
 Revocation is irreversible. Suppose the validator learns of revocation by UCAN CID. In that case, the UCAN and all of its derivatives in such a cache MUST be marked as invalid, and all validations immediately fail without needing to walk the entire structure.
 
-## 5.3 Replay Attack Prevention
+## Replay Attack Prevention
 
 Replay attack prevention is REQUIRED. Every UCAN token MUST have a unique CID. Some simple strategies for implementing uniqueness tracking include maintaining a set of previously seen CIDs, or requiring that nonces be monotonically increasing per principal. This MAY be the same structure as a validated UCAN memoization table (if one is implemented).
 
 Maintaining a secondary token expiry index is RECOMMENDED. This enables garbage collection and more efficient search. In cases of very large stores, normal cache performance techniques MAY be used, such as Bloom filters, multi-level caches, and so on.
 
-## 5.4 Beyond Single System Image
+## Beyond Single System Image
 
 > As we continue to increase the number of globally connected devices, we must embrace a design that considers every single member in the system as the primary site for the data that it is generates. It is completely impractical that we can look at a single, or a small number, of globally distributed data centers as the primary site for all global information that we desire to perform computations with.
 >
@@ -438,7 +438,7 @@ sequenceDiagram
     Bob ->> Alice: invoke(CRDT_ID, merge, {"Banana", "Carrot"}, proof: [➋])
 ```
 
-## 5.5 Wrapping Existing Systems
+## Wrapping Existing Systems
 
 In the RECOMMENDED scenario, the agent controlling a resource has a unique reference to it. This is always possible in a system that has adopted capabilities end-to-end.
 
@@ -477,25 +477,25 @@ sequenceDiagram
     DBAgent ->>- Bob: ACK
 ```
 
-# 6. FAQ
+# FAQ
 
-## 6.1 What prevents an unauthorized party from using an intercepted UCAN?
+## What prevents an unauthorized party from using an intercepted UCAN?
 
 UCANs always contain information about the sender and receiver. A UCAN is signed by the sender (the `iss` field DID) and can only be created by an agent in possession of the relevant private key. The recipient (the `aud` field DID) is required to check that the field matches their DID. These two checks together secure the certificate against use by an unauthorized party.
 
-## 6.2 What prevents replay attacks on the invocation use case?
+## What prevents replay attacks on the invocation use case?
 
 A UCAN delegated for purposes of immediate invocation MUST be unique. If many requests are to be made in quick succession, a nonce can be used. The receiving agent (the one to perform the invocation) checks the hash of the UCAN against a local store of unexpired UCAN hashes.
 
 This is not a concern when simply delegating since presumably the recipient agent already has that UCAN.
 
-## 6.3 Is UCAN secure against person-in-the-middle attacks?
+## Is UCAN secure against person-in-the-middle attacks?
 
 _UCAN does not have any special protection against person-in-the-middle (PITM) attacks._
 
 If a PITM attack was successfully performed on a UCAN delegation, the proof chain would contain the attacker's DID(s). It is possible to detect this scenario and revoke the relevant UCAN but this does require special inspection of the topmost `iss` field to check if it is the expected DID. Therefore, it is strongly RECOMMENDED to only delegate UCANs to agents that are both trusted and authenticated and over secure channels.
 
-# 7. Related Work and Prior Art
+# Related Work and Prior Art
 
 [SPKI/SDSI] is closely related to UCAN. A different format is used, and some details vary (such as a delegation-locking bit), but the core idea and general usage pattern are very close. UCAN can be seen as making these ideas more palatable to a modern audience and adding a few features such as content IDs that were less widespread at the time SPKI/SDSI were written.
 
@@ -511,7 +511,7 @@ If a PITM attack was successfully performed on a UCAN delegation, the proof chai
 
 [Verifiable credentials] are a solution for data about people or organizations. However, they are aimed at a slightly different problem: asserting attributes about the holder of a DID, including things like work history, age, and membership.
 
-# 8. Acknowledgments
+# Acknowledgments
 
 Thank you to [Brendan O'Brien] for real-world feedback, technical collaboration, and implementing the first Golang UCAN library.
 
@@ -545,10 +545,10 @@ We want to especially recognize [Mark Miller] for his numerous contributions to 
 
 <!-- Internal Links -->
 
-[Command]: #33-command
-[overcoming SSI]: #54-beyond-single-system-image
+[Command]: #command
+[overcoming SSI]: #beyond-single-system-image
 [sub-specifications]: #sub-specifications
-[wrapping existing systems]: #55-wrapping-existing-systems
+[wrapping existing systems]: #wrapping-existing-systems
 
 <!-- External Links -->
 
--- a/README.md
+++ b/README.md
@@ -491,7 +491,7 @@ This is not a concern when simply delegating since presumably the recipient agen
 
 _UCAN does not have any special protection against person-in-the-middle (PITM) attacks._
 
-Were a PITM attack successfully performed on a UCAN delegation, the proof chain would contain the attacker's DID(s). It is possible to detect this scenario and revoke the relevant UCAN but does require special inspection of the topmost `iss` field to check if it is the expected DID. Therefore, it is strongly RECOMMENDED to only delegate UCANs to agents that are both trusted and authenticated and over secure channels.
+If a PITM attack was successfully performed on a UCAN delegation, the proof chain would contain the attacker's DID(s). It is possible to detect this scenario and revoke the relevant UCAN but this does require special inspection of the topmost `iss` field to check if it is the expected DID. Therefore, it is strongly RECOMMENDED to only delegate UCANs to agents that are both trusted and authenticated and over secure channels.
 
 # 7. Related Work and Prior Art
 
--- a/README.md
+++ b/README.md
@@ -228,7 +228,7 @@ There are several roles that an agent MAY assume:
 | Agent     | The general class of entities and principals that interact with a UCAN                           |
 | Audience  | The Principal delegated to in the current UCAN. Listed in the `aud` field                        |
 | Executor  | The Agent that actually performs the action described in an invocation                           |
-| Invoker   | A Principal that requests that an Executor perform some action that uses the Invoker's authority |
+| Invoker   | A Principal that requests an Executor perform some action that uses the Invoker's authority |
 | Issuer    | The Principal of the current UCAN. Listed in the `iss` field                                     |
 | Owner     | A Subject that controls some external resource                                                    |
 | Principal | An agent identified by DID (listed in a UCAN's `iss` or `aud` field)                             |
--- a/README.md
+++ b/README.md
@@ -283,7 +283,7 @@ A resource is some data or process that can be uniquely identified by a [URI]. I
 
 A resource describes the noun of a capability. The resource pointer MUST be provided in [URI] format. Arbitrary and custom URIs MAY be used, provided that the intended recipient can decode the URI. The URI is merely a unique identifier to describe the pointer to — and within — a resource.
 
-Having a unique agent represent a resource (and act as its manager) is RECOMMENDED. However, to help traditional ACL-based systems transition to certificate capabilities, an agent MAY manage multiple resources, and act as the registrant in the ACL system. These situations can be represented as follows:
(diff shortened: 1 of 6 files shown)


Gozala (changes requested):
At the high level it looks good and improves on some pain points we've experienced. That said, reading this:

1. I'm no longer sure what the structure of the UCAN looks like any more perhaps even one example somewhere would help.
2. I think imposing certain structure on commands / abilities would really simplify things yet will not prevent anyone to go with a different approach if they choose so, they'll only loose ability to leverage certain tooling that will become possible.
3. Given significant changes here, I would be more comfortable calling this 0.20 or something and let it settle before we mark it 1.0 or 1.0-rc
4. Document often refers to proof chains, yet leaves a lot of ambiguity what they are. I find that confusing, UCANs without proofs feel more like certificates that you can chain into authorization. I feel like it would be good to either call that out explicitly or avoid referring to proof chains

README.md
```diff
@@ -85,872 +118,382 @@ Unlike many authorization systems where a service controls access to resources i
...
+    CRDT -->> Bob: delegate(CRDT_ID, merge)
+    
+    Note over Bob, Carol: Bob Invites Carol
+    Bob -->> Carol: delegate(CRDT_ID, merge)
 
-  ... --> *
+    Note over Alice, Carol: Direct P2P Gossip
+    Carol ->> Bob: invoke(CRDT_ID, merge, {"Carrot"}, proof: [➋,➍])
+    Alice ->> Carol: invoke(CRDT_ID, merge, {"Apple"}}, proof: [➊])
+    Bob ->> Alice: invoke(CRDT_ID, merge, {"Banana", "Carrot"}, proof: [➋])
```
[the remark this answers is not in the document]
expede: @Gozala Right, I'm trying to show a different use case. You're treating it as a physical location, as opposed to a replicated agent

README.md
```diff
@@ -85,872 +118,382 @@ Unlike many authorization systems where a service controls access to resources i
...
+Ultimately this comes down to a question of push vs pull. In push, the subject MUST be the specific site being pushed to ("I command you to apply the following updates to your state").
 
-Bob would like to share access to write into a shared directory with Alice. Normally, either Bob would have to be aware of all of Alice's public keys, or the device that Bob delegates to would have to manually redelegate to Alice's other devices. With `ucan:*`, Bob can delegate to `did:key:aliceRoot`, and `did:key:alicePhone` can use the `ucan:*` resource to access Bob's shared directory.
+Pull is the broad class of situations where an Invoker doesn't require that a particular replica apply its state. Applying a change to a local CRDT replica and maintaining a UCAN invocation log is a valid update to "the CRDT": a version of the CRDT Subject exists locally even if the Subject's private key is not present. Gossiping these changes among agents allows each to apply changes that it becomes aware of. Thanks to the invocation log (or equivalent integrated directly into the CRDT), provenance of authority is made transparent.
```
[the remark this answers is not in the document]
expede: It's the difference between a "single system image" (AKA "location") versus a replicated agent. Take a WNFS or an Automerge document for example: there are two things that you may want to express:

1. I can write into this data staructure in general (even locally)
2. I'm allowed to update some specific replica (location)

(1) isn't in reference to any particular replica. I can make the update directly on my node, and attach the proofs for anyone to inspect if they **pull** the CRDT.

In (2), you **push** to a certain replica. You're consuming storage on a particular node, and works much more like a traditional HTTP PUT or POST request.

README.md
```diff
@@ -85,872 +118,382 @@ Unlike many authorization systems where a service controls access to resources i
...
+Revocation is irreversible. Suppose the validator learns of revocation by UCAN CID or issuer DID. In that case, the UCAN and all of its derivatives in such a cache MUST be marked as invalid, and all validations immediately fail without needing to walk the entire structure.
 
-The `ucan` URI scheme defines URI selectors for UCANs.
+## 5.3 Replay Attack Prevention
 
-``` abnf
-ucan = "ucan:" ["//" resource-owner-did "/"] ucan-selector
-ucan-selector = "*" / uri-scheme / ucan-cid
-```
+Replay attack prevention is REQUIRED. Every UCAN token MUST have a unique CID. Some simple strategies for implementing uniqueness tracking include maintaining a set of previously seen CIDs, or requiring that nonces be monotonically increasing per principal. This MAY be the same structure as a validated UCAN memoization table (if one is implemented).
```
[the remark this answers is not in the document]
expede: >  Recommendation seems to be never issue the same UCAN which seems odd

Yeah, you should never use the same Invocation twice. They MUST be unique.

If they're not unique, then a PITM can intercept and replay the token. As a simple example for intuition: if your invocation says "increment counter" they can just run it thousand times even though the one submitting it doens't control the issuer's DID.

README.md
```diff
@@ -85,872 +118,382 @@ Unlike many authorization systems where a service controls access to resources i
...
+This store SHOULD be indexed by CID (content addressing). Multiple indices built on top of this store MAY be used to improve capability search or selection performance.
 
-# 4. Reserved Resources
+## 5.2 Memoized Validation
 
-The following resources are REQUIRED to be implemented.
+Aside from revocation, capability validation is idempotent. Marking a CID (or capability index inside that CID) as valid acts as memoization, obviating the need to check the entire structure on every validation. This extends to distinct UCANs that share a proof: if the proof was previously reviewed and is not revoked, it is RECOMMENDED to consider it valid immediately.
 
-## 4.1 `ucan`
+Revocation is irreversible. Suppose the validator learns of revocation by UCAN CID or issuer DID. In that case, the UCAN and all of its derivatives in such a cache MUST be marked as invalid, and all validations immediately fail without needing to walk the entire structure.
```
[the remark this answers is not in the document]
expede: Yeah that's true. This is from the old revocation format. Fixed.

README.md
```diff
@@ -85,872 +118,382 @@ Unlike many authorization systems where a service controls access to resources i
...
+A validator MAY keep a local store of UCANs that it has received. UCANs are immutable but also time-bound so that this store MAY evict expired or revoked UCANs.
 
-For more on this representation, please refer to [canonical collections].
+This store SHOULD be indexed by CID (content addressing). Multiple indices built on top of this store MAY be used to improve capability search or selection performance.
 
-# 4. Reserved Resources
+## 5.2 Memoized Validation
 
-The following resources are REQUIRED to be implemented.
+Aside from revocation, capability validation is idempotent. Marking a CID (or capability index inside that CID) as valid acts as memoization, obviating the need to check the entire structure on every validation. This extends to distinct UCANs that share a proof: if the proof was previously reviewed and is not revoked, it is RECOMMENDED to consider it valid immediately.
```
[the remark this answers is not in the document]
expede: > UCAN may include multiple capabilities 

Not anymore... they're single resource

README.md
```diff
@@ -85,872 +118,382 @@ Unlike many authorization systems where a service controls access to resources i
...
+|---------|------------------------------------------------------------|
+| Subject | `did:key:z6MkhaXgBZDvotDkL5257faiztiGiC2QtKLGpbnnEGta2doK` |
+| Ability | `msg/send`                                                 |
+| Caveat  | `{sender: "mailto:alice@example.com", "day": "friday"}`    |
 
-```json
-"https://example.com",
-"https://example.com/blog"
-```
+## 3.7 Authority
```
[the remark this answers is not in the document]
expede: Authority is the noun — the act of "delegating authority" is "authorization"

README.md
```diff
@@ -85,872 +118,382 @@ Unlike many authorization systems where a service controls access to resources i
...
+|-----------|--------------------------------------------------------------------------------------------------|
+| Agent     | The general class of entities and principals that interact with a UCAN                           |
+| Audience  | The Principal delegated to in the current UCAN. Listed in the `aud` field                        |
+| Executor  | The Agent that actually performs the action described in an invocation                           |
+| Invoker   | A Principal that requests that an Executor perform some action that uses the Invoker's authority |
+| Issuer    | The Principal of the current UCAN. Listed in the `iss` field                                     |
+| Owner     | A Subject that controls some external resource                                                    |
+| Principal | An agent identified by DID (listed in a UCAN's `iss` or `aud` field)                             |
+| Revoker   | The Issuer listed in a proof chain that revokes a UCAN                                           |
+| Subject   | The Principal who's authority is delegated or invoked                                            |
```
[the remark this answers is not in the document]
expede: Subject is a standard field in many auth tokens

README.md
```diff
@@ -85,872 +118,382 @@ Unlike many authorization systems where a service controls access to resources i
...
-}
+``` mermaid
+sequenceDiagram
+    Alice -->> Bob: Delegate
+    Bob ->> Bob: Validate
+    Bob -->> Carol: Delegate
+    Carol ->> Carol: Validate
+    Carol ->> Alice: Invoke
+    Alice ->> Alice: Validate
+    Alice ->> Alice: Execute
```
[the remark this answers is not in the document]
expede: You should be able to check the UCAN chain at each hop, no?

README.md
```diff
@@ -1,63 +1,96 @@
...
 
 ## 1.4 Inversion of Control
 
-Unlike many authorization systems where a service controls access to resources in their care, location-independent, offline, and leaderless resources require control to live with the user. Therefore, the same data MAY be used across many applications, data stores, and users.
+[Inversion of control] is achieved due to two properties: self-certifying delegation and reference passing. There is no Authorization Server (AS) that sits between requestors and resources. In traditional terms, the owner of a UCAN resource is the resource server (RS) directly.
```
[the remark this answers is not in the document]
expede: I can add words here to say that the user is the RS... it's mainly for people coming from OAuth.

FWIW Fission used to do it the way you described, and we've come to reguard that as a mistake 😅

README.md
```diff
@@ -85,872 +118,382 @@ Unlike many authorization systems where a service controls access to resources i
...
+Commands MUST NOT be case-sensitive. There MUST be at least one path segment as a namespace. For example, `http/put` and `db/put` MUST be treated as unique from each other.
 
-### 3.2.5 Facts
+### 3.4.1 Ability
 
-The OPTIONAL `fct` field contains a map of arbitrary facts and proofs of knowledge. The enclosed data MUST be self-evident and externally verifiable. It MAY include information such as hash preimages, server challenges, a Merkle proof, dictionary data, etc.
+Abilities abstract over [Command]s to allow for extension of UCAN delegations.
 
-#### Examples
+Abilities MAY be organized in a hierarchy. A typical example is a superuser capability ("anything") on a file system. Another is read vs write access, such that in an HTTP context, `WRITE` implies `PUT`, `PATCH`, `DELETE`, and so on. Organizing abilities this way allows for adding more options over time in a backward-compatible manner, avoiding the need to reissue UCANs with new resource semantics.
```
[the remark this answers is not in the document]
expede: > I would highly recommend organizing hierarchies through name-spacing and direct reader to doin the same.

We've considered this, but it breaks when you want to be able to update these later. As described, you have to impose these fully up front syntactically. Validation already has to understand the proxying pattern (`ucan/*`) to connect disparate chains, so validtaion is not fully syntactically driven

README.md
```diff
@@ -85,872 +118,382 @@ Unlike many authorization systems where a service controls access to resources i
...
-``` json
-"nnc": "1701-D"
-```
+Commands MUST NOT be case-sensitive. There MUST be at least one path segment as a namespace. For example, `http/put` and `db/put` MUST be treated as unique from each other.
 
-### 3.2.5 Facts
+### 3.4.1 Ability
 
-The OPTIONAL `fct` field contains a map of arbitrary facts and proofs of knowledge. The enclosed data MUST be self-evident and externally verifiable. It MAY include information such as hash preimages, server challenges, a Merkle proof, dictionary data, etc.
+Abilities abstract over [Command]s to allow for extension of UCAN delegations.
```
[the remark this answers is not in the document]
expede: Yeah this definitely needs to be tightened up. The difference is that a Command is something that can be called in an Invocation. You can't really call arguments on `http/*`, and adding any args to `http/*` make it much more difficult to delegate portion (`http/* -> http/put`) because it's really difficult to figure out how one would check the delegation.

I've been thinking about maybe just dropping the distinction and saying "too bad, you need to trace the `http/put` all the way to the root, you can't use `http/*` to prove it" but that seems like such a shame.

README.md
```diff
@@ -1,63 +1,96 @@
...
 
-This signature chain is the root of trust. Private keys themselves SHOULD NOT move from one context to another: this is what the delegation mechanism provides: "sharing authority without sharing keys."
+To achieve these properties, object capabilities have two requirements: [fail-safe], and locality preservation. The emphasis on consistency rules out partition tolerance[^pcec].
 
-UCANs (and other forms of PKI) depend on the ambient authority of the owner of each resource. This means that the discharging agent must be able to verify the root ownership at decision time. The rest of the chain in-between is self-certifying.
+## Security Considerations
+
+Each UCAN includes an assertions of what it is allowed to do. "Proofs" are positive evidence (elsewhere called "witnesses") of the possession of rights. They are cryptographically verifiable chains showing that the UCAN issuer either claims to directly own a resource, or that it was delegated to them by some claimed owner. In the most common case, the root owner's ID is the only globally unique identity for the resource.
```
bgins, line 75:
```suggestion
Each UCAN includes assertions of what it is allowed to do. "Proofs" are positive evidence (elsewhere called "witnesses") of the possession of rights. They are cryptographically verifiable chains showing that the UCAN issuer either claims to directly own a resource, or that it was delegated to them by some claimed owner. In the most common case, the root owner's ID is the only globally unique identity for the resource.
```
"an" looks unneeded here.

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
+The Subject and Command fields are REQUIRED. Any non-normative extensions are OPTIONAL.
 
-# 4. Reserved Resources
+For example, a capability may used to represent the ability to send email from a certain address to others at `@example.com`.
 
-The following resources are REQUIRED to be implemented.
+| Field      | Example                                                                                      |
+|------------|----------------------------------------------------------------------------------------------|
+| Subject    | `did:key:z6MkhaXgBZDvotDkL5257faiztiGiC2QtKLGpbnnEGta2doK`                                   |
+| Command    | `msg/send`                                                                                   |
```
Gozala, line 350: Since we have moved to paths in delegation

```suggestion
| Command    | `/msg/send`                                                                                  |
```

Gozala (approved):
Just added a minor correction in the example command

README.md
```diff
@@ -1,63 +1,97 @@
...
 
-Many high-value applications run in hostile environments. In recognition of this, many vendors now include public key functionality, such as [non-extractable keys in browsers][browser api crypto key], [certificate systems for external keys][fido], and [secure hardware enclaves] in widespread consumer devices.
+Two related models that work exceptionally well in the above context are Simple Public Key Infrastructure ([SPKI][spki rfc]) and object capabilities ([OCAP]). Since offline operation and self-verifiability are two requirements, UCAN adopts a [certificate capability model] related to [SPKI].
 
-Two related models that work exceptionally well in the above context are Simple Public Key Infrastructure ([SPKI][spki rfc]) and object capabilities ([OCAP]). Since offline operation and self-verifiability are two requirements, UCAN adopts an approach closely related to SPKI. UCANs follow the "capabilities as certificates" model, with extensions for revocation and stateful capabilities.
+## Intuition for Auth System Differences
```
bumblefudge, line 54:
```suggestion
## Intuition for Auth System Differences
```

README.md
```diff
@@ -1,63 +1,97 @@
...
+To achieve these properties, object capabilities have two requirements: [fail-safe], and locality preservation. The emphasis on consistency rules out partition tolerance[^pcec].
 
-This signature chain is the root of trust. Private keys themselves SHOULD NOT move from one context to another: this is what the delegation mechanism provides: "sharing authority without sharing keys."
+## Security Considerations
 
-UCANs (and other forms of PKI) depend on the ambient authority of the owner of each resource. This means that the discharging agent must be able to verify the root ownership at decision time. The rest of the chain in-between is self-certifying.
+Each UCAN includes an assertions of what it is allowed to do. "Proofs" are positive evidence (elsewhere called "witnesses") of the possession of rights. They are cryptographically verifiable chains showing that the UCAN issuer either claims to directly own a resource, or that it was delegated to them by some claimed owner. In the most common case, the root owner's ID is the only globally unique identity for the resource.
```
bumblefudge, line 76: As [gozala mentioned a while ago](https://github.com/ucan-wg/spec/pull/173#pullrequestreview-1683120805), I think the concept of "proof chain" can be a bit of a weaselword for a first-time reader.  I would recommend adding a ### Proof Chains section defining what you do and don't mean by the term, immediately before ## Security Considerations and after ### Object Capabilities, since the ## Intuition section seems like the right place to set up novel primitives before returning to the macro-view in ## Security Considerations.

README.md
```diff
@@ -1,63 +1,97 @@
...
+
+Root capability issuers function as verifiable, distributed roots of trust. The delegation chain is by definition a provenance log. Private keys themselves SHOULD NOT move from one context to another. Keeping keys unique to each physical device and unique per use case is RECOMMENDED to reduce opportunity for keys to leak, and limit blast radius in the case of compromises. "Sharing authority without sharing keys" is provided by capabilities, so there is no reason to share keys directly.
```
bumblefudge, line 78: The first sentence has always struck me as confusing, and I was afraid to say anything for fear of sounding dumb until now.  Is the root _capability_ itself the root of trust, or is the ID by which its issuer is known the root of trust? either way it feels like you're saying that a root capability should be trusted anywhere its issuer's ID is trusted, and can be verified/refreshed/checked-in-on...

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
 └─────────────┘   └─────────────┘   └─────────────┘
 ```
 
-# 2. Terminology
+This additionally allows UCAN to model auth for [eventually consistent and replicated state][overcoming SSI].
```
bumblefudge, line 122:
```suggestion
This additionally allows UCAN to model authorization for [eventually consistent and replicated state], thus [overcoming SSI] constraints.
```

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
-| --------- | ----------- |
-| Agent     | The general class of entities and principals that interact with a UCAN |
-| Validator | Any agent that interprets a UCAN to determine that it is valid, and which capabilities it grants |
-| Principal | An agent identified by DID (listed in a UCAN's `iss` or `aud` field) |
-| Audience  | The principal delegated to in the current UCAN. Listed in the `aud` field |
-| Signer    | A principal that can sign payloads |
-| Issuer    | The signer of the current UCAN. Listed in the `iss` field |
-| Revoker   | The issuer listed in a proof chain that revokes a UCAN |
-| Owner     | The root issuer of a capability, who has some proof that they fully control the resource |
+| Name      | Description                                                                                      |
```
bumblefudge, line 128:
```suggestion
| Name      | Description                                                                                      |
```
linting like crazy over here

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
 └─────────────┘   └─────────────┘   └─────────────┘
 ```
 
-# 2. Terminology
+This additionally allows UCAN to model auth for [eventually consistent and replicated state][overcoming SSI].
 
-## 2.1 Roles
+# Roles
```
bumblefudge, line 124:
```suggestion
## Roles
```
I think single-hash renames the file in some forms of markdown? in any case my markdown linter plugin discourages using them within a document

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
-```
-
-## 2.6 Authority
+A resource describes the noun of a capability. The resource pointer MUST be provided in [URI] format. Arbitrary and custom URIs MAY be used, provided that the intended recipient can decode the URI. The URI is merely a unique identifier to describe the pointer to — and within — a resource.
 
-The set of capabilities delegated by a UCAN is called its "authority." This functions as a declarative description of delegated abilities.
+Having a unique agent represent a resource (and act as its manager) is RECOMMENDED. However, to help traditional ACL-based systems transition to certificate capabilities, an agent MAY manage multiple resources, and [act as the registrant in the ACL system][wrapping existing systems].
 
-Merging capability authorities MUST follow set semantics, where the result includes all capabilities from the input authorities. Since broader capabilities automatically include narrower ones, this process is always additive. Capability authorities can be combined in any order, with the result always being at least as broad as each of the original authorities.
+Unless explicitly stated, the Resource of a UCAN MUST be the Subject.
```
bumblefudge, line 189: I don't understand where and how any other state of affairs could be explicitly stated? Do you mean 

```suggestion
Unless all expected recipients of a UCAN will understand how the Subject of a UCAN relates to the Resource being authorized because this relationship is explicitly clarified out-of-band, it can be assumed by all parties that a UCAN's subject is coterminous with the Resource authorized by that UCAN.
```
?  I assume not, but if you asked _me_ to make that sentence more verbose, this is what I would come up with, which feels like a very wrong guess.

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
-| Revoker   | The issuer listed in a proof chain that revokes a UCAN |
-| Owner     | The root issuer of a capability, who has some proof that they fully control the resource |
+| Name      | Description                                                                                      | 
+|-----------|--------------------------------------------------------------------------------------------------|
+| Agent     | The general class of entities and principals that interact with a UCAN                           |
+| Audience  | The Principal delegated to in the current UCAN. Listed in the `aud` field                        |
+| Executor  | The Agent that actually performs the action described in an invocation                           |
+| Invoker   | A Principal that requests an Executor perform some action that uses the Invoker's authority |
+| Issuer    | The Principal of the current UCAN. Listed in the `iss` field                                     |
+| Owner     | A Subject that controls some external resource                                                    |
```
bumblefudge, line 135:
```suggestion
| Owner     | A Subject that controls some external resource                                                   |
```

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
-| Signer    | A principal that can sign payloads |
-| Issuer    | The signer of the current UCAN. Listed in the `iss` field |
-| Revoker   | The issuer listed in a proof chain that revokes a UCAN |
-| Owner     | The root issuer of a capability, who has some proof that they fully control the resource |
+| Name      | Description                                                                                      | 
+|-----------|--------------------------------------------------------------------------------------------------|
+| Agent     | The general class of entities and principals that interact with a UCAN                           |
+| Audience  | The Principal delegated to in the current UCAN. Listed in the `aud` field                        |
+| Executor  | The Agent that actually performs the action described in an invocation                           |
+| Invoker   | A Principal that requests an Executor perform some action that uses the Invoker's authority |
```
bumblefudge, line 133:
```suggestion
| Invoker   | A Principal that requests an Executor perform some action that uses the Invoker's authority      |
```

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
-└──────────────────┬─────────────────┘
-                   │
-
-               AxY U BxZ
-                 authority
-```
+## Issuer & Audience
 
-The capability authority is the total rights of the authorization space down to the relevant volume of authorizations. Individual capabilities MAY overlap; the authority is the union. Except for [rights amplification], every unique delegation MUST have equal or narrower capabilities from their delegator. Inside this content space, you can draw a boundary around some resource(s) (their type, identifiers, and paths or children) and their capabilities.
+The Issuer (`iss`) and Audience (`aud`) can be conceptualized as the sender and receiver (respectively) of a postal letter. Every UCAN MUST be signed with the private key associated with the DID in the `iss` field.
```
bumblefudge, line 193:
```suggestion
The Issuer (`iss`) and Audience (`aud`) can be conceptualized as the sender and receiver (respectively) of a postal letter. Every UCAN MUST be signed with the private key associated with the DID in the `iss` field. 

Note that the UCAN system assumes a 1:1 relationship between DID URLs and public keys, and all examples are given using `did:key`, where bare DIDs are resolvable directly and deterministically to single keys. For DID methods that support multiple keys for a given DID, verification-specific [DID URLs using fragment identifiers](https://www.w3.org/TR/did-core/#fragment) MUST be used for `iss` and `aud` values to preserve interoperability; non-key-based `verificationMethod`s MUST NOT be used, for the same reason.
```
I think I promised this as a PR over a year ago, I'm realizing as I write this up now 🙈

design/cryptosuite.md
```diff
@@ -0,0 +1,17 @@
+# Cryptosuite
+
+## Signature Algorithms
+
+In an ideal world, a single algorithm (Ed25519) would be available everywhere. For many historical reasons, P-256 and `secp256k1` are the only practical options for many applications, such as the [WebCrypto API] and many cryptocurrency wallets. A design goal of UCAN is to leverage existing public key infrastructure (PKI), and supporting these three algorithms achieves that goal. Requiring multiple signature types for broad interoperability is not an unusual design choice; for example, [JWT] supports multiple signature algorithms for a similar reason.
+
+The NIST report on post-quantum computing will also be released in the next few months, so we anticipate a shift to new algorithms in the medium-term.
```
bumblefudge, line 7: Which report do you mean? "The" report came out in 2016, but I assume you mean a more definitive one?
https://csrc.nist.gov/Projects/post-quantum-cryptography/publications

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
-  "bafkreihogico5an3e2xy3fykalfwxxry7itbhfcgq6f47sif6d7w6uk2ze": "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsInVhdiI6IjAuMS4wIn0.eyJhdWQiOiJkaWQ6a2V5OnpTdEVacHpTTXRUdDlrMnZzemd2Q3dGNGZMUVFTeUExNVc1QVE0ejNBUjZCeDRlRko1Y3JKRmJ1R3hLbWJtYTQiLCJpc3MiOiJkaWQ6a2V5Ono1QzRmdVAyRERKQ2hoTUJDd0FrcFlVTXVKWmROV1dINU5lWWpVeVk4YnRZZnpEaDNhSHdUNXBpY0hyOVR0anEiLCJuYmYiOjE1ODg3MTM2MjIsImV4cCI6MTU4OTAwMDAwMCwic2NwIjoiLyIsInB0YyI6IkFQUEVORCIsInByZiI6bnVsbH0.Ay8C5ajYWHxtD8y0msla5IJ8VFffTHgVq448Hlr818JtNaTUzNIwFiuutEMECGTy69hV9Xu9bxGxTe0TpC7AzV34p0wSFax075mC3w9JYB8yqck_MEBg_dZ1xlJCfDve60AHseKPtbr2emp6hZVfTpQGZzusstimAxyYPrQUWv9wqTFmin0Ls-loAWamleUZoE1Tarlp_0h9SeV614RfRTC0e3x_VP9Ra_84JhJHZ7kiLf44TnyPl_9AbzuMdDwCvu-zXjd_jMlDyYcuwamJ15XqrgykLOm0WTREgr_sNLVciXBXd6EQ-Zh2L7hd38noJm1P_MIr9_EDRWAhoRLXPQ"
-}
-```
+A capability is the association of an ability to a subject: `subject x command x policy`.
 
-For more on this representation, please refer to [canonical collections].
+The Subject and Command fields are REQUIRED. Any non-normative extensions are OPTIONAL.
```
bumblefudge, line 343: are all policies "non-normative extensions"? are any other fields?  OPTIONAL for implementations to handle?  Dropped if unknown?  I know this is all stated in the delegation spec, but just speaking as a reader, a tiny bit more detail, although redundant, might be helpful here for high-level readers!

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
 
+### Reserved Commands
 
-### 5.2.1 Bottom
-
-In concept there is a "bottom" ability ("none" or "void"), but it is not possible to represent in an ability. As it is merely the absence of any ability, it is not possible to construct a capability with a bottom ability.
+#### `/ucan` Namespace
 
-# 6. Validation
+The `/ucan` Command namespace MUST be reserved. This MUST include any ability string matching the regex `^ucan\/.*`. This is important for keeping a space for community-blessed Commands in the future, such as standard library Commands, such as [Revocation].
```
bumblefudge, line 456:
```suggestion
The `/ucan` Command namespace MUST be reserved. This MUST include any ability string matching the regex `^ucan(\/(^\m\/)+)+`. This is important for keeping a space for community-blessed Commands in the future, such as standard library Commands, such as [Revocation].
```
I'm no regex wizard but I'm guessing this regex is older than the prohibition on trailing `/`s, which this newer, uglier regex bans.

Also, I couldn't tell if `/ucan` was supposed to be a valid command (or a valid blank-check for future stdlib commands), but if so, that last `+` (1 or more) should be a `*` (0 or more).

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
-
-  executor -. accesses .-> resource
-  rootCap -. references .-> resource
-
-  subgraph root [Root UCAN]
-    rootIss(iss: Alice)
-    rootAud(aud: Bob)
-    rootCap("cap: (Storage, crud/*)")
-  end
+This field SHOULD NOT be used to sign arbitrary data, such as signature challenges. See the [`meta`][Metadata] field for more.
```
bumblefudge, line 478:
```suggestion
This field SHOULD NOT be used to sign arbitrary data, such as signature challenges. See the [`meta`][#Metadata] field for more.
```

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
+#### `/ucan` Namespace
 
-# 6. Validation
+The `/ucan` Command namespace MUST be reserved. This MUST include any ability string matching the regex `^ucan\/.*`. This is important for keeping a space for community-blessed Commands in the future, such as standard library Commands, such as [Revocation].
 
-Each capability has its own semantics, which needs to be interpretable by the target resource handler. Therefore, a validator SHOULD NOT reject UCANs with resources that it does not know how to interpret.
+## Attenuation
 
-If any of the following criteria are not met, the UCAN MUST be considered invalid.
+Attenuation is the process of constraining the capabilities in a delegation chain. Each direct delegation MUST either directly restate or attenuate (diminish) its capabilities.
```
bumblefudge, line 460:
```suggestion
Attenuation is the process of constraining the capabilities in a delegation chain. Each direct delegation MUST either explicitly restate or attenuate (diminish) its capabilities.
```

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
+Token resolution is transport specific. The exact format is left to the relevant UCAN transport specification. At minimum, such a specification MUST define at least the following:
 
-All proofs MUST contain time bounds equal to or broader than the UCAN being delegated. If the proof expires before the outer UCAN — or starts after it — the reader MUST treat the UCAN as invalid. Delegation inside of the time bound is called "timely delegation." These conditions MUST hold even if the current wall clock time is inside of incorrectly delegated bounds. 
+1. Request protocol
+2. Response protocol
+3. Collections format
 
-A UCAN is valid inclusive from the `nbf` time and until the `exp` field. If the current time is outside of these bounds, the UCAN MUST be considered invalid. When setting these bounds, a delegator or invoker SHOULD account for expected clock drift. Use of time bounds this way is called "timely invocation."
+Note that if an instance cannot dereference a CID at runtime, the UCAN MUST fail validation. This is consistent with the [constructive semantics] of UCAN.
```
bumblefudge, line 470: Note that this is the first mention of a CID in the entire high-level spec, without a normref or definition.

Editorial nitpicking aside, I think it might be clearer to refer to them as `link`s in this and other section (which, in IPLD, are effectively just IPFS CIDs) rather than refer to them as CIDs, because what matters is that they are identifiers they can't be defererenced, not anything about their nature as _content_ identifiers or multiformats, etc etc.  (I.e., I can imagine an RDF/JSON-LD version of this where the same link-dereferenceability rule applies to DID URLs, or HTTPS URLS, or some other kinds of dereferenceable URIs...)

bumblefudge: For example:
```suggestion
Note that if a UCAN includes links in the form of [IPFS CIDs][CID] to subgraphs, and a parsing implementation cannot expand the full graph because one or more of these links cannot be dereferenced at runtime, then the UCAN MUST fail validation. This is consistent with the [constructive semantics] of UCAN.
```

README.md
```diff
@@ -1020,70 +789,100 @@ Were a PITM attack successfully performed on a UCAN delegation, the proof chain
...
-[`did:ion`]: https://github.com/decentralized-identity/ion
+[`base58btc`]: https://github.com/multiformats/multibase/blob/master/multibase.csv#L21
 [`did:key`]: https://w3c-ccg.github.io/did-method-key/
-[base32]: https://github.com/multiformats/multibase/blob/master/multibase.csv#L12
+[`secp256k1`]: https://en.bitcoin.it/wiki/Secp256k1
 [browser api crypto key]: https://developer.mozilla.org/en-US/docs/Web/API/CryptoKey
-[canonical collections]: #71-canonical-json-collection
 [capabilities]: https://en.wikipedia.org/wiki/Object-capability_model
 [caps as keys]: http://www.erights.org/elib/capability/duals/myths.html#caps-as-keys
+[certificate capability model]: http://wiki.erights.org/wiki/Capability-based_Active_Invocation_Certificates
```
bumblefudge, line 868:
```suggestion
[certificate capability model]: http://wiki.erights.org/wiki/Capability-based_Active_Invocation_Certificates
[CID]: https://docs.ipfs.tech/concepts/content-addressing/
```
Unless the uglier, raw github definition seems more appropriate or canonical for whatever reason? 
https://github.com/multiformats/cid

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
+}
 ```
 
-In the above diagram, Alice has some storage. This storage may exist in one location with a single source of truth, but to help build intuition this example is location independent: local versions and remote stored copies are eventually consistent, and there is no one "correct" copy. As such, we list the owner (Alice) directly on the resource.
+# Metadata
 
-Alice delegates access to Bob. Bob then redelegates to Carol. Carol invokes the UCAN as part of a REST request to a compute service. To do this, she MUST both provide proof that she has access (the UCAN chain), and MUST delegate access to the discharging compute service. The discharging service MUST check that the root issuer (Alice) is in fact the owner (typically the creator) of the resource. This MAY be listed directly on the resource, as it is here. Once the UCAN chain and root ownership are validated, the storage service performs the write.
+The OPTIONAL `meta` field contains a map of arbitrary metadata, facts, and proofs of knowledge. The enclosed data MUST be self-evident and externally verifiable. It MAY include information such as hash preimages, server challenges, a Merkle proof, dictionary data, etc.
```
bumblefudge, line 491:
```suggestion
The OPTIONAL `meta` field contains a map of arbitrary metadata, facts, and proofs of knowledge. The enclosed data SHOULD be self-evident and externally verifiable. It MAY include information such as hash preimages, server challenges, a Merkle proof, dictionary data, etc.
```
It feels like this MUST is untestable, and is more like an instruction to UCAN-generating/encoding implementations than anything that could be included in a conformance test? It might be better phrased as a consuming behavior, i.e., "Validators MUST NOT consider the contents when proofing UCAN invocation".

corner-case: do CIDs in metadata have to be dereferenceable at runtime as well? or does their being unmeaningful trump their being IPLD 😄

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
-Delegation is the act of granting another principal (the delegate) the capability to use a resource that another has (the delegator). A constructive "proof" acts as the authorization proof for a delegation. Proofs are either the owning principal's signature or a UCAN with access to that capability in its authority.
-
-Each direct delegation leaves the ability at the same level or diminishes it. The only exception is in "rights amplification," where a delegation MAY be proven by one-or-more proofs of different types if part of the resource's semantics. 
-
-Note that delegation is a separate concept from [invocation]. Delegation is the act of granting a capability to another, not the act of using it (invocation), which has additional requirements.
-
-## 2.8 Attenuation
-
-Attenuation is the process of constraining the capabilities in a delegation chain.
+Please see the [Cryptosuite] section for more detail on DIDs.
```
bumblefudge, line 202: Narrator: but there was no more detail on DIDs there.

bumblefudge:
```suggestion
Please see the [Cryptosuite](#cryptosuite) section for more detail on DIDs.
```

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
 {
-  "aud": "did:key:zH3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV",
-  "iss": "did:key:zAKJP3f7BD6W4iWEQ9jwndVTCBq8ua2Utt8EEjJ6Vxsf",
   // ...
+  "meta": {
+    "challenges": {
+      "example.com": "abcdef",
+      "another.example.net": "12345"
+    },
+    "sha3_256": {
```
bumblefudge, line 505: not `SHA-256`, i.e. SHA2? or is that deliberate?

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
+| Parameter    | REQUIRED Configuration |
+|--------------|------------------------|
+| Version      | [CIDv1]                | 
+| [Multibase]  | [`base58btc`]          | 
+| [Multihash]  | [SHA-256]              |
+| [Multicodec] | [DAG-CBOR]             |
 
-A UCAN token MUST be referenced as a [base32] [CIDv1]. [SHA2-256] is the RECOMMENDED hash algorithm.
+> [!NOTE]
+> All CIDs encoded as above start with the characters `zdpu`.
```
bumblefudge, line 540: If they're canonically DAG-CBOR why is the multibase needed, as opposed to [staying in binary](https://github.com/multiformats/cid?tab=readme-ov-file#how-does-it-work)? Maybe I've been talking to the wrong people but I thought CIDv1 was _canonically_ binary and _optionally_ multibased, so it feels like an extra roundtrip to multibase it and then drop back down to CBOR...

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
+| Hash      | [SHA-256]                         |                                      |
+| Signature | [Ed25519], [P-256], [`secp256k1`] | Preference of Ed25519 is RECOMMENDED |
+| [DID]     | [`did:key`]                       |                                      |
 
-Each capability MUST either be originated by the issuer (root capability, or "parenthood") or have one-or-more proofs in the `prf` field to attest that this issuer is authorized to use that capability ("introduction"). In the introduction case, this check MUST be recursively applied to its proofs until a root proof is found (i.e. issued by the resource owner).
+## Encoding
 
-Except for rights amplification (below), each capability delegation MUST have equal or narrower capabilities from its proofs. The time bounds MUST also be equal to or contained inside the time bounds of the proof's time bounds. This lowering of rights at each delegation is called "attenuation."
+All UCANs MUST be canonically encoded with [DAG-CBOR] for signing. A UCAN MAY be presented or stored in other [IPLD] formats (such as [DAG-JSON]), but converted to DAG-CBOR for signature validation.
```
bumblefudge, line 526: I feel like the ` { "/" : { "bytes": "...` notation needs to be explained a little here, rather than assuming the reader both wants to and can reasonably click that "DAG-CBOR" link and come back 10 minutes later knowing what IPLD is, if they don't already... because if _I_ am confused, who is your target reader 😏

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
+| [Multibase]  | [`base58btc`]          | 
+| [Multihash]  | [SHA-256]              |
+| [Multicodec] | [DAG-CBOR]             |
 
-A UCAN token MUST be referenced as a [base32] [CIDv1]. [SHA2-256] is the RECOMMENDED hash algorithm.
+> [!NOTE]
+> All CIDs encoded as above start with the characters `zdpu`.
 
-The [`0x55` raw data][raw data multicodec] codec MUST be supported. If other codecs are used (such as [`0x0129` `dag-json` multicodec][dag-json multicodec]), the UCAN MUST be able to be interpreted as a valid JWT (including the signature).
+The resolution of these addresses is left to the implementation and end-user, and MAY (non-exclusively) include the following: local store, a distributed hash table (DHT), gossip network, or RESTful service.
```
bumblefudge, line 542: Redundant, sure, but here might be a good place to remind the reader that a validator MUST fail verification of a UCAN with subgraphs linked by CID that do not dereference at time of consumption of invocations.

README.md
bumblefudge: also, are they "Arguments"  now? I am seeing links elsewhere in this document to `[Arguments]` but there is no `### Arguments` section (or for that matter a `### Policy` or `### Extensions` section, either!), so probably time to add one.

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
+A validator MAY keep a local store of UCANs that it has received. UCANs are immutable but also time-bound so that this store MAY evict expired or revoked UCANs.
 
-This format makes it easy to select the relevant UCAN, confirm that the issuer is in the proof chain for some or all capabilities, and validate that the revocation was signed by the same issuer.
+This store SHOULD be indexed by CID (content addressing). Multiple indices built on top of this store MAY be used to improve capability search or selection performance.
 
-Any other proofs in the selected UCAN not issued by the same DID as the revocation issuer MUST be treated as valid.
+## Memoized Validation
 
-Revocations MAY be deleted once the UCAN that they reference expires or otherwise becomes invalid via its proactive mechanisms.
+Aside from revocation, capability validation is idempotent. Marking a CID (or capability index inside that CID) as valid acts as memoization, obviating the need to check the entire structure on every validation. This extends to distinct UCANs that share a proof: if the proof was previously reviewed and is not revoked, it is RECOMMENDED to consider it valid immediately.
```
bumblefudge, line 610: Should that by "Aside from revocation **and expiration**,..." ? I am also confused by the corner case mentioned here by Gozala: 

Let's say Alice delegates to Bob a capability `/crud/` over a hard disk, then Bob sends Carol (me) a capability `/crud/write/`.  Before I go to invoke it at Alice's hard disk, Alice's original parent cap expired or was revoked.  Bob sends me a new sub-delegation identical except for having a fresher parent.  That's a different token, right? This idempotency stuff doesn't apply because if it's a different chain, it's a different token?

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
 
-### 6.6.1 Example
+Revocation is irreversible. Suppose the validator learns of revocation by UCAN CID. In that case, the UCAN and all of its derivatives in such a cache MUST be marked as invalid, and all validations immediately fail without needing to walk the entire structure.
```
bumblefudge, line 612: I'd love for "Suppose the validator learns of revocation by UCAN CID" to be expanded a bit, I'm having trouble supposing from that very high-level summary 😅

bumblefudge:
Forgive the billion little comments, few of them are worth discussing in further detail, mostly just idle suggestions.  Is merging this blocked by anything? Just waiting for invocation to go to v1rc1?

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
+| Field   | Type                                      | Required | Description                                                 |
+|---------|-------------------------------------------|----------|-------------------------------------------------------------|
+| `iss`   | `DID`                                     | Yes      | Issuer DID (sender)                                         |
+| `aud`   | `DID`                                     | Yes      | Audience DID (receiver)                                     |
+| `sub`   | `DID`                                     | Yes      | Principal that the chain is about (the [Subject])           |
+| `cmd`   | `String`                                  | Yes      | The [Command] to eventually invoke                          |
+| `args`  | `{String : Any}`                          | Yes      | Any [Arguments] that MUST be present in the Invocation      |
+| `nonce` | `Bytes`                                   | Yes      | Nonce                                                       |
+| `meta`  | `{String : Any}`                          | No       | [Meta] (asserted, signed data) — is not delegated authority |
```
bumblefudge, line 596: Maybe break out the non-required fields into a separate table below, since all of this is under MUST CONTAIN?

And throw `iat` in there, too?

expede: The rationale is that `iat` is not in the delegation spec since it's not particularly helpful for delegation. Invocation does have an `iat`, so it's not a common field across all UCAN formats

README.md
expede: This is an example of putting in arbitrary data. It literally doesn't matter

README.md
```diff
@@ -85,932 +119,667 @@ Unlike many authorization systems where a service controls access to resources i
...
-  subgraph del2 [Final UCAN]
-    del2Iss(iss: Carol) --> del1Aud
-    del2Aud(aud: Compute Service)
-    del2Cap("cap: (Storage, crud/*)") --> del1Cap
-  end
+``` js
+{
+  // ...
+  "nonce": {"/": {"bytes": "bGlnaHQgd29yay4"}}
+}
```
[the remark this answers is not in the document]
expede: > Reading quickly I got thrown off by the "/" and thought it was an IPLD thing.

it is an IPLD thing. This is how DAG-JSON represents binary data

PR #196, opened 2025-12-02
chore: add note about empty segments

Gozala:
I ran into issues caused by decode/encode being lossy, which was cause by cmd serialization as mentioned here https://github.com/ucan-wg/delegation/issues/26. Spec is ambiguous in regards to empty segments so I'm proposing to amend it to be explicit about it

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,7 +402,7 @@ While arbitrary semantics MAY be described, they MUST apply to the target resour
 
 ### Segment Structure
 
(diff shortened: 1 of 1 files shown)


alanshaw (approved):

Gozala:
@expede mind giving this an approval and/or making @alanshaw a code owner so his approvals suffice branch protections

PR #200, opened 2026-02-10
fix: regression from #192

Gozala:
I have landed #192 without realizing and was a syntax error rendering policy.json invalid. This  fixes to what I think was intended json


alanshaw (approved):

Gozala:
@expede do you ming giving this an approval and/or making @alanshaw a code owner so his approvals are sufficient

PR #202, opened 2026-02-11
fix: invocation proof order

### LINKED ISSUE #201: Issues with invocation fixtures
I ran into couple of issues when exercising integration with invocation spec in rs-ucan that could use some clarification or revision on fixtures

1. [Order of proofs in the invocation](https://github.com/ucan-wg/invocation/issues/41) assumes `invoker <- subject` where's at some implementations seem to assume `subject → invoker`. In my current workstream I'm basically resorting to accepting either order, but it would be nice to have spec and test vectors align.

2. Some of the invocation fixtures error with `UnauthorizedSubject` in rs-ucan instead of expected `Expired` / `TooEarly` errors. Specifically, "expired proof", "inactive proof" and "expired invocation". That is because proof authorizes sub `did:key:z6MkmT9j...` but then invocation claims sub `did:key:z6MkmJce...` which is misaligned.

   I think if subs were aligned rs-ucan would also fail on time bounds, it's just we asses subject alignment earlier. Either way I think fixing fixtures so subs align would be a good idea.

3. I think invocation fixtures should also include time field denoting time of validation, otherwise things that are too early eventually may become not anymore early. Less likely to see same for already expired ones. Either way I think it's a good idea to to have time captured so test can be deterministic and don't have time variable in them.

alanshaw:
resolves https://github.com/ucan-wg/spec/issues/201

This PR fixes the proof order in valid invocation fixtures. It also adds a `time` field - a Unix timestamp in seconds at which the validation should occur.

refs https://github.com/alanshaw/ucantone/pull/2

### DIFF
--- a/fixtures/1.0.0/invocation.json
+++ b/fixtures/1.0.0/invocation.json
@@ -1,5 +1,5 @@
 {
-  "comments": "Encoded as dag-json.",
+  "comments": "Encoded as dag-json. Time field is the time at which the validation should occur - a Unix timestamp in seconds.",
   "invalid": [
     {
       "description": "it has no proofs",
@@ -12,7 +12,8 @@
         }
       },
       "name": "no proof",
-      "proofs": []
+      "proofs": [],
+      "time": 1767225600
     },
     {
       "description": "a proof is not provided or resolvable externally",
@@ -25,7 +26,8 @@
         }
       },
       "name": "missing proof",
-      "proofs": []
+      "proofs": [],
+      "time": 1767225600
     },
     {
       "description": "a proof is expired",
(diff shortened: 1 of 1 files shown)


Gozala:
I'll integrate this into rs-ucan to verify it works as expected and provide feedback based on findings (I'm not good at parsing cbor in base64 encoded with my eyes 😅)

Gozala (approved):
Looks good and rust implementation is in compliance with it now 🎉

Thanks @alanshaw

Gozala:
@alanshaw I think branch protection rules prevent merges unless commits are signed, you may want to update your [git config so it signs commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) I'll amend this one to unblock and merge

Gozala:
> @alanshaw I think branch protection rules prevent merges unless commits are signed, you may want to update your [git config so it signs commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) I'll amend this one to unblock and merge

Actually I can not do that because PR is from your repo and I don't have commit rights to it, so I think you may need to do this

alanshaw:
@Gozala all good now?

alanshaw:
@Gozala ping

